PR #6, opened 2022-10-09
Add CI

SapiensAnatis:
GitHub will now build the code and test it whenever you push or open a pull request


SapiensAnatis:
need to fix testfixture so that the test pipeline passes

PR #16, opened 2022-10-22
Lots of summon things, also Update_Flags

SteelGram:
Literally cannot list all this stuff


DragaliaAPI.Test/Integration/Dragalia/SummonTest.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            ApiContext apiContext = scope.ServiceProvider.GetRequiredService<ApiContext>();
+            int size =
+                data.result_unit_list[0].entity_type == 7
+                    ? apiContext.PlayerDragonReliability.Count()
+                        + apiContext.PlayerDragonData.Count()
+                    : apiContext.PlayerCharaData.Count();
+            _output.WriteLine(size.ToString());
+        }
+    }
```
SapiensAnatis, line 137: I don't really like these tests that just output to stdout instead of containing asserts. It might be more meaningful to test that the database has been updated with the UpdateDataList contents

DragaliaAPI/Models/Data/DewValues.cs
```diff
@@ -4,17 +4,17 @@ namespace DragaliaAPI.Models.Data;
...
-    RARITY_3 = 150,
-    RARITY_5_STORY = 300,
-    RARITY_4 = 2200,
-    RARITY_5 = 8500
+    Rarity4Story = 100,
+    Rarity3 = 150,
+    Rarity5Story = 300,
+    Rarity4 = 2200,
+    Rarity5 = 8500
 }
```
SapiensAnatis, line 18: Should this and the other eldwater values not be constants? As far as I can tell you just immediately put them into a Dictionary<int, int> so there is no value in having a type for it

DragaliaAPI/Models/Data/ManaNodes.cs
```diff
@@ -48,7 +48,7 @@ public static ManaNodes SetManaCircleNodesFromSet(IEnumerable<int> enumerable)
 
     public static SortedSet<int> GetSetFromManaNodes(ManaNodes flag)
     {
-        SortedSet<int> nodes = new();
+        SortedSet<int> nodes = new SortedSet<int>();
```
SapiensAnatis, line 51: what's wrong with the implicit `new()`?

DragaliaAPI/Models/Database/Savefile/DbPlayerUserData.cs
```diff
@@ -76,6 +77,13 @@ public class DbPlayerUserData : IDbHasAccountId
 
     public int TutorialFlag { get; set; }
 
+    [NotMapped]
+    public ISet<int> TutorialFlagList
+    {
+        get => TutorialFlagUtil.ConvertIntToFlagIntList(TutorialFlag);
+        set => TutorialFlag = TutorialFlagUtil.ConvertFlagIntListToInt(value);
+    }
+
```
SapiensAnatis, line 86: ideally the repository should do conversions like this, that way it's more testable

DragaliaAPI/Services/SummonService.cs
```diff
@@ -10,21 +19,94 @@ public class SummonService : ISummonService
...
         _dragonDataService = dragonDataService;
     }
 
-    public List<SummonEntity> GenerateSummonResult(int numSummons)
+    public record BannerSummonInfo(
+        Dictionary<EntityTypes, SummonableEntity> featured,
+        Dictionary<EntityTypes, SummonableEntity> normal,
+        double baseSsrRate,
+        double baseRRate
+    );
```
SapiensAnatis, line 40: I tend to prefer declaring types in their own file rather than in services/controllers. I guess it just makes them easier to find?

DragaliaAPI/Services/SavefileWriteService.cs
```diff
@@ -1,70 +1,155 @@
...
     }
 
     public async Task<UpdateDataList> CommitSummonResult(
-        List<SummonEntity> summonResult,
+        List<SimpleSummonReward> summonResult,
         string deviceAccountId,
         bool giveDew = true
     )
     {
         int dewGained = 0;
```
SapiensAnatis, line 37: not adding eldwater to savefile?

DragaliaAPI/Models/Dragalia/Responses/TutorialUpdateFlagsResponse.cs
```diff
@@ -0,0 +1,12 @@
+﻿using DragaliaAPI.Models.Dragalia.Responses.Common;
+using DragaliaAPI.Models.Dragalia.Responses.UpdateData;
+using MessagePack;
+
+namespace DragaliaAPI.Models.Dragalia.Responses;
+
+[MessagePackObject(keyAsPropertyName: true)]
+public record TutorialUpdateStepResponse(TutorialUpdateStepData data)
```
SapiensAnatis, line 8: you've mixed up the filename for the different tutorial response models

DragaliaAPI/Models/Data/ManaNodes.cs
SapiensAnatis: actually this may have been a change I did which was then not reflected in your branch, so it looks like you reversed it

DragaliaAPI.Test/Integration/Dragalia/SummonTest.cs
SteelGram: the 10x does, can probably add it to these as we fix them in general. The output is very handy to quickly see the results in the Test output

DragaliaAPI/Services/SavefileWriteService.cs
SteelGram: probably a remnant of removing it from the commit and moving it to the summoncontroller method

DragaliaAPI/Services/SummonService.cs
SteelGram: I was working on figuring out proper rolling when we decided to put it as polish, this was port of it. It is only used in the unused method for figuring out odds and probably should eventually it should go into its own file when it contains everything the method needs to calculate them

PR #46, opened 2022-12-21
DragonController (Draft)

SteelGram:
Upgrading & Bonding with dragons


DragaliaAPI.Database/Entities/DbPlayerStoryState.cs
```diff
@@ -20,7 +20,7 @@ public class DbPlayerStoryState : IDbHasAccountId
     [Required]
     public int StoryId { get; set; }
 
-    [Column("State")]
+    [Column("StoryState")]
```
SapiensAnatis, line 23: Why has this changed?

DragaliaAPI.Database/Repositories/IInventoryRepository.cs
```diff
@@ -11,4 +11,7 @@ public interface IInventoryRepository : IBaseRepository
     DbPlayerMaterial AddMaterial(string deviceAccountId, Materials type);
     Task<DbPlayerMaterial?> GetMaterial(string deviceAccountId, Materials materialId);
     IQueryable<DbPlayerMaterial> GetMaterials(string deviceAccountId);
+    DbPlayerDragonGift AddDragonGift(string deviceAccountId, DragonGifts type);
+    Task<DbPlayerDragonGift?> GetDragonGift(string deviceAccountId, DragonGifts materialId);
```
SapiensAnatis, line 15: Why is the gifts method an IQueryable but the singular Gift method a Task?

IMO it is preferable to return an IQueryable so that the caller can customize the query before executing it. But whichever way you decide to do it they should be consistent

DragaliaAPI.Database/Utils/DragonConstants.cs
```diff
@@ -0,0 +1,287 @@
...
+
+public static class DragonConstants
+{
+    public static byte GetMaxLevelFor(int rarity, int limitBreak)
+    {
+        int baseLevel;
+        switch (rarity)
+        {
+            case 3:
+                baseLevel = MinMaxLevel[rarity - 3] + (10 * Math.Min(4, limitBreak));
```
SapiensAnatis, line 14: is the min/clamp logic because 5UB doesn't increase level?

DragaliaAPI.Database/Utils/DragonConstants.cs
```diff
@@ -0,0 +1,287 @@
...
+        DragonGifts.GoldenChalice,
+        DragonGifts.JuicyMeat,
+        DragonGifts.Kaleidoscope,
+        DragonGifts.FloralCirclet,
+        DragonGifts.CompellingBook,
+        DragonGifts.ManaEssence,
+        DragonGifts.GoldenChalice
+    };
+
+    public readonly static Dictionary<DragonGifts, int> favorVals = new Dictionary<
```
SapiensAnatis, line 175: should probably be immutable dictionaries and the arrays immutable too

DragaliaAPI.Shared/Definitions/Enums/Materials.cs
```diff
@@ -549,6 +549,61 @@ public static class UpgradeMaterials
         { Materials.SucculentDragonfruit, 3500 }
     };
 
+    public readonly static Dictionary<int, int> dragonBuildupXp = new Dictionary<int, int>()
```
SapiensAnatis, line 552: These data structures are handy but I'm not sure the enums folder is the best place for them

DragaliaAPI/Controllers/Dragalia/DragonController.cs
```diff
@@ -0,0 +1,510 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+using static DragaliaAPI.Database.Utils.DragonConstants;
+
+namespace DragaliaAPI.Controllers.Dragalia;
+
+[Route("dragon")]
+[Consumes("application/octet-stream")]
+[Produces("application/octet-stream")]
+[ApiController]
```
SapiensAnatis, line 23: these are not needed when inheriting from DragaliaControllerBase (at least on the latest branches)

DragaliaAPI/Controllers/Dragalia/DragonController.cs
```diff
@@ -0,0 +1,510 @@
...
+        this.inventoryRepository = inventoryRepository;
+        this.storyRepository = storyRepository;
+        this.updateDataService = updateDataService;
+        this.mapper = mapper;
+        _dragonDataService = charaDataService;
+        this.dragonService = dragonService;
+    }
+
+    [Route("buildup")]
+    [HttpPost]
```
SapiensAnatis, line 57:
```suggestion
    [HttpPost("buildup")]
```

DragaliaAPI/Controllers/Dragalia/DragonController.cs
```diff
@@ -0,0 +1,510 @@
...
+                    matIds.Contains(dbMat.MaterialId)
+                    || dbMat.MaterialId == Materials.FortifyingDragonscale
+                    || dbMat.MaterialId == Materials.AmplifyingDragonscale
+            )
+            .ToDictionaryAsync(dbMat => dbMat.MaterialId);
+        foreach (GrowMaterialList mat in request.grow_material_list)
+        {
+            if (mat.quantity < 0)
+            {
+                throw new ArgumentException("Invalid quantity for MaterialList");
```
SapiensAnatis, line 78: You should now throw DragaliaExceptions with an appropriate resultcode which will return a 200 with a bad status_code in data_headers and show a message to the user, should they ever manage to trigger this

DragaliaAPI/Controllers/Dragalia/DragonController.cs
```diff
@@ -0,0 +1,510 @@
...
+                        break;
+                    case Materials.FortifyingDragonscale:
+                        playerDragonData.HpPlusCount = (byte)
+                            Math.Min(
+                                playerDragonData.HpPlusCount + materialList.quantity,
+                                DragonConstants.MaxHpEnhance
+                            );
+                        break;
+                    default:
+                        throw new ArgumentException("Invalid MaterialList");
```
SapiensAnatis, line 190: Doesn't this duplicate material validation logic in the first part of the controller? My preference would be to keep this one if so

DragaliaAPI/Controllers/Dragalia/DragonController.cs
```diff
@@ -0,0 +1,510 @@
...
+                    playerDragonData.Exp
+                        + UpgradeMaterials.dragonBuildupXp[
+                            _dragonDataService.GetData((int)playerDragonData.DragonId).Rarity
+                        ],
+                    DragonConstants.XpLimits[maxLevel - 1]
+                );
+                DbPlayerDragonData dragonSacrifice = unitRepository
+                    .GetAllDragonData(DeviceAccountId)
+                    .Where(x => x.DragonKeyId == materialList.id)
+                    .First();
```
SapiensAnatis, line 210: Should be FirstAsync. And method needs to be changed to async Task for this

DragaliaAPI/Controllers/DragaliaControllerBase.cs
```diff
@@ -21,7 +21,17 @@ public override OkObjectResult Ok(object? value)
...
+
+    public OkObjectResult Ok(ResultCode resultCode, object? value)
+    {
+        return base.Ok(
+            new DragaliaResponse<object>(
+                value ?? throw new ArgumentNullException(nameof(value)),
+                resultCode
             )
         );
     }
```
SapiensAnatis, line 37: as mentioned before this is now achieved via DragaliaExceptions, so this overload shouldn't exist -- idk if you added it

DragaliaAPI/Services/UpdateDataService.cs
```diff
@@ -29,9 +29,36 @@ public UpdateDataList GetUpdateDataList(string deviceAccountId)
...
+            .Select(x => x.Quantity)
+            .FirstOrDefaultAsync();
+        t2.Wait();
+
+        DiamondData diaData = new DiamondData()
+        {
+            free_diamond = (int)t.Result,
+            paid_diamond = (int)t2.Result
+        };
+
```
SapiensAnatis, line 57: The UpdateDataService should not be performing queries and waiting for them. The point is to query the change tracker, not the actual database itself. This implementation will return diamond_data on every single update_data_list

DragaliaAPI/Services/UpdateDataService.cs
```diff
@@ -29,9 +29,36 @@ public UpdateDataList GetUpdateDataList(string deviceAccountId)
...
         return new()
```
SapiensAnatis, line 58: I think you forgot to add dragon gifts to this?

DragaliaAPI.Database/Migrations/20230207152134_develop_7.cs
```diff
@@ -0,0 +1,79 @@
...
+                table: "PlayerDragonGift");
+
+            migrationBuilder.DropColumn(
+                name: "OwnerAccountId",
+                table: "PlayerDragonGift");
+
+            migrationBuilder.RenameColumn(
+                name: "StoryState",
+                table: "PlayerStoryState",
+                newName: "State");
```
SapiensAnatis, line 28: is it possible to remove this entirely from the migrations? i.e squash the ones you have together?

also your migrations aren't following the new naming scheme based on the version -- i will soon merge in 20230205020012_1-1-0_alpha_6 so could you go with 1-1-0_alpha_7 please

DragaliaAPI.Database/Repositories/InventoryRepository.cs
```diff
@@ -172,4 +172,52 @@ public async Task<bool> CheckQuantity(IEnumerable<KeyValuePair<Materials, int>>
...
+            {
+                dbGift.Quantity = 1;
+            }
+            else
+            {
+                dbGift = AddDragonGift(deviceAccountId, gift);
+                dbGift.Quantity = 1;
+            }
+        }
+        return true;
```
SapiensAnatis, line 221: why does this return a bool if it will never return false?

DragaliaAPI.Shared/MasterAsset/MasterAsset.cs
```diff
@@ -29,4 +29,14 @@ public static class MasterAsset
     private static readonly Lazy<MasterAssetData<WeaponBodies, WeaponBody>> _WeaponBody =
         new(() => new("WeaponBody.json", x => x.Id));
     public static MasterAssetData<WeaponBodies, WeaponBody> WeaponBody => _WeaponBody.Value;
+    private static readonly Lazy<MasterAssetData<int, StoryData>> _DragonStories =
```
SapiensAnatis, line 32: i think i said to use ints as the key but I didn't realize they were split -- is it still not possible to use the enum because of non-playable units?

DragaliaAPI.Shared/Resources/Export.json
```diff
@@ -0,0 +1,10962 @@
+[
```
SapiensAnatis, line 1: why do you have CharaStories.json, DragonStories.json, Export.json, and two view jsons? surely you only need the first 2

DragaliaAPI.Database/Entities/DbPlayerCharaData.cs
```diff
@@ -34,8 +34,43 @@ public class DbPlayerCharaData : IDbHasAccountId
     [Column("Level")]
     public byte Level { get; set; } = 1;
 
-    [Column("AddMaxLevel")]
-    public byte AdditionalMaxLevel { get; set; } = 0;
+    [NotMapped]
+    public byte AdditionalMaxLevel
```
SapiensAnatis, line 38: could you add a short comment explaining how this works, I think you told me but I already forgot :upside_down_face:

DragaliaAPI/Controllers/Dragalia/LoadController.cs
```diff
@@ -66,14 +66,17 @@ public async Task<DragaliaResult> Index()
...
+                    .Select(this.mapper.Map<DragonGiftList>),
                 talisman_list = savefile.TalismanList.Select(this.mapper.Map<TalismanList>),
                 weapon_body_list = savefile.WeaponBodyList.Select(this.mapper.Map<WeaponBodyList>),
                 party_list = savefile.PartyList.Select(this.mapper.Map<PartyList>),
                 quest_story_list = savefile.StoryStates
                     .Where(x => x.StoryType == StoryTypes.Quest)
                     .Select(mapper.Map<QuestStoryList>),
                 unit_story_list = savefile.StoryStates
-                    .Where(x => x.StoryType == StoryTypes.Chara)
+                    .Where(x => x.StoryType == StoryTypes.Chara || x.StoryType == StoryTypes.Dragon)
```
SapiensAnatis, line 79: dragon isn't an actual story type, both chara and dragon are all stored in unit_story_list so we should probably rework the enum

DragaliaAPI/Controllers/DragaliaControllerBase.cs
```diff
@@ -33,6 +33,16 @@ public override OkObjectResult Ok(object? value)
...
+    public OkObjectResult Ok(ResultCode resultCode, object? value)
+    {
+        return base.Ok(
+            new DragaliaResponse<object>(
+                value ?? throw new ArgumentNullException(nameof(value)),
+                resultCode
+            )
+        );
+    }
+
```
SapiensAnatis, line 45: not sure if this was added by you or just a leftover from rebase, but it is not needed, could you remove it? We have Code() for returning non-1 ResultCodes which are always errors and have no other data with them

DragaliaAPI/Models/Generated/Components.cs
```diff
@@ -1787,14 +1787,16 @@ public AtgenDragonBonus(UnitElement elemental_type, float dragon_bonus, float hp
 [MessagePackObject(true)]
 public class AtgenDragonGiftRewardList
 {
-    public int dragon_gift_id { get; set; }
-    public int is_favorite { get; set; }
+    public DragonGifts dragon_gift_id { get; set; }
+
+    [MessagePackFormatter(typeof(BoolToIntFormatter))]
```
SapiensAnatis, line 1792: if you want to import these then you also need to add the json formatter

DragaliaAPI/Services/DragonService.cs
```diff
@@ -0,0 +1,578 @@
...
+                    }
+                    await storyRepository.GetOrCreateStory(
+                        deviceAccountId,
+                        StoryTypes.Dragon,
+                        dragonStories[nextStoryUnlockIndex]
+                    );
+                    reward.is_release_story = 1;
+                    if (levelIndex == 3)
+                    {
+                        //TODO: Add Epithet to account
```
SapiensAnatis, line 220: you only get the epithet when reading the story no?

DragaliaAPI.Database/Repositories/InventoryRepository.cs
```diff
@@ -42,7 +42,7 @@ public DbPlayerCurrency AddCurrency(string deviceAccountId, CurrencyTypes type)
     public async Task<DbPlayerCurrency?> GetCurrency(string deviceAccountId, CurrencyTypes type)
     {
         return await this.apiContext.PlayerWallet.FirstOrDefaultAsync(
-            entry => entry.CurrencyType == type
+            entry => entry.DeviceAccountId == deviceAccountId && entry.CurrencyType == type
         );
```
SapiensAnatis, line 46: please use findasync here

DragaliaAPI.Database/Entities/DbPlayerDragonGift.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.EntityFrameworkCore;
+
+namespace DragaliaAPI.Database.Entities;
+
+[Table("PlayerDragonGift")]
+[Index(nameof(DeviceAccountId))]
+public class DbPlayerDragonGift : IDbHasAccountId
+{
+    [Column("DeviceAccountId")]
+    [Required]
```
SapiensAnatis, line 13: fyi this doesn't do anything to the database, the column is always NOT NULL unless the type is nullable, i think this just causes a runtime exception if it is not specified

DragaliaAPI.Database/Entities/DbPlayerDragonReliability.cs
```diff
@@ -53,5 +54,11 @@ public static DbPlayerDragonReliability Create(string deviceAccountId, Dragons i
             GetTime = DateTimeOffset.UtcNow,
             LastContactTime = DateTimeOffset.UtcNow
         };
+        if (DragonConstants.fullBondDragons.Contains(id))
+        {
+            newReliability.Exp = DragonConstants.bondXpLimits[^1];
+            newReliability.Level = (byte)DragonConstants.bondXpLimits.Length;
+        }
+        return newReliability;
```
SapiensAnatis, line 62: Can you change this to use the masterasset DefaultReliabilityLevel and remove the fullbonddragonslist

DragaliaAPI/Controllers/Dragalia/CharaController.cs
```diff
@@ -286,6 +289,12 @@ public async Task<DragaliaResult> Buildup([FromBody] CharaBuildupRequest request
         DbPlayerMaterial upgradeMat =
             await inventoryRepository.GetMaterial(DeviceAccountId, mat)
             ?? inventoryRepository.AddMaterial(DeviceAccountId, mat);
+        DbPlayerCurrency playerCurrency =
+            await inventoryRepository.GetCurrency(DeviceAccountId, CurrencyTypes.Rupies)
+            ?? throw new DragaliaException(
+                ResultCode.CommonMaterialShort,
+                "Insufficient Rupies for reset"
+            );
```
SapiensAnatis, line 297: not needed

SapiensAnatis (changes requested):
I think this is almost ready!

Along with the above comments could you also add some integration tests -- having done the unit tests there's no need to cover each possible case but just a simulated request to each endpoint would be good

SapiensAnatis:
Just ran some code coverage in my IDE (will probably look at setting this up so everyone can view coverage) and it looks like you are missing tests for:

- giving gifts to the puppy
- adding dragon augments
- unbinding dragons with spheres

DragaliaAPI/Controllers/Nintendo/AnalyticsControllers.cs
```diff
@@ -2,17 +2,16 @@
 
 namespace DragaliaAPI.Controllers.Nintendo;
```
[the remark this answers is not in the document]
SteelGram: I'd rather not remove it because I highly dislike your dependency on the baas and think it wouldn't hurt to keep the server complete and able to handle these requests without it.
That being said I don't even remember uncommenting this, must've been a long time ago. I'll just re-comment it

DragaliaAPI/Services/DragonService.cs
```diff
@@ -0,0 +1,1040 @@
...
+        ];
+        Dictionary<DragonGifts, DbPlayerDragonGift> gifts = await inventoryRepository
+            .GetDragonGifts(deviceAccountId)
+            .Where(
+                x =>
+                    x.DragonGiftId == DragonGifts.FreshBread
+                    || x.DragonGiftId == DragonGifts.TastyMilk
+                    || x.DragonGiftId == DragonGifts.StrawberryTart
+                    || x.DragonGiftId == DragonGifts.HeartyStew
+                    || x.DragonGiftId == rotatingGift
```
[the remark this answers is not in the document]
SteelGram: you mean a column for whether it's a purchasable or an inventory gift?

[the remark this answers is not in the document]
SapiensAnatis: doesn't have to be a column if it's just a getter, it's just that you've repeated this logic in the updatedataservice and one other place I can't remember

DragaliaAPI.Test/Integration/Dragalia/DragonControllerTest.cs
```diff
@@ -0,0 +1,578 @@
+﻿using DragaliaAPI.Controllers.Dragalia;
```
SapiensAnatis, line 1: Can you rename this file/class to DragonTest to keep with the naming scheme of integration tests? xxxControllerTest is normally used for unit tests of controllers

DragaliaAPI.Test/Integration/Dragalia/DragonControllerTest.cs
```diff
@@ -0,0 +1,578 @@
...
+        };
+
+        DragonSetLockData? response = (
+            await client.PostMsgpack<DragonSetLockData>("dragon/reset_plus_count", request)
+        ).data;
+
+        response.Should().NotBeNull();
+        response.update_data_list.Should().NotBeNull();
+        response.update_data_list.dragon_list.Should().NotBeNullOrEmpty();
+        DragonList returnDragon = response.update_data_list.dragon_list
```
SapiensAnatis, line 212: You don't have to fix this, but FYI the preceding 3 assertions are not needed because this statement would throw an NullReference/InvalidOperation exception if they were untrue and the test would fail anyway

DragaliaAPI.Test/Integration/Dragalia/DragonControllerTest.cs
```diff
@@ -0,0 +1,578 @@
...
+            .Where(x => (long)x.dragon_key_id == dragon.DragonKeyId)
+            .First();
+        returnDragon.attack_plus_count.Should().Be(0);
+        response.update_data_list.user_data.Should().NotBeNull();
+        response.update_data_list.user_data.coin.Should().Be(startCoin - (20000 * 50));
```
SapiensAnatis, line 217: Can you check that the user receives their augments back?

DragaliaAPI.Test/Integration/Dragalia/DragonControllerTest.cs
```diff
@@ -0,0 +1,578 @@
...
+            .First()
+            .is_buy.Should()
+            .Be(0);
+
+        response.dragon_gift_reward_list.Should().NotBeNullOrEmpty();
+        DragonReliabilityList dragonData = response.update_data_list.dragon_reliability_list
+            .Where(x => x.dragon_id == Dragons.HighChthonius)
+            .First();
+        dragonData.reliability_total_exp.Should().Be(400);
+        dragonData.reliability_level.Should().Be(3);
```
SapiensAnatis, line 281: Can you check that the user's rupies are deducted correctly?

DragaliaAPI/Services/DragonService.cs
```diff
@@ -0,0 +1,1032 @@
...
+
+        //DbPlayerCurrency playerCurrency =
+        //    await inventoryRepository.GetCurrency(deviceAccountId, CurrencyTypes.Rupies)
+        //    ?? throw new DragaliaException(
+        //        ResultCode.CommonMaterialShort,
+        //        "Insufficient Rupies for reset"
+        //    );
+        DbPlayerUserData userData = await userDataRepository.LookupUserData();
+        int cost =
+            20000
```
SapiensAnatis, line 795: can you make this a const instead of having it as a magic number

SapiensAnatis (commented):
Aside from these nitpicks, it's pretty much there! Can look to release this in the next day or so I think. Just need to do the save import stuff on my end

DragaliaAPI/Models/Generated/Components.cs
SapiensAnatis: This is not the imported model, oops

SapiensAnatis (approved):

PR #57, opened 2022-12-26
Implement BaaS login

SapiensAnatis:
Switch old auth flow to new auth flow based on LukeFZ's external BaaS server. This enables login information to be synchronized with other servers and accounts to be linked.

- We no longer handle Nintendo endpoints, thus NintendoLoginController, DeviceAccountService, and DbDeviceAccount are obsolete **except** for local testing, where a config switch means that these are still used.
- Now we will receive a valid JWT at /tool/auth which we can validate and decode to get a BaaS account ID. This will be the new primary key of all of the tables. As a result, savefile creation will occur at /tool/auth if the PlayerUserData table contains no entries for the account ID. (This is not done at /tool/signup because the client will not call that if it has switched from another server emulator).
- This is tested by mocking the API call and replacing it with a custom keypair and tokens generated with that. The tests should not be calling the external service.


SapiensAnatis:
Config needs updating with config changes in #46

PR #68, opened 2022-12-31
Fixed clear time and mission completion tracking

Nightmerp:


DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
```diff
@@ -38,9 +38,25 @@ IUpdateDataService updateDataService
...
     {
         DungeonSession session = await this.dungeonService.FinishDungeon(request.dungeon_key);
+        DbQuest? oldQuestData = await this.questRepository
+            .GetQuests(this.DeviceAccountId)
+            .SingleOrDefaultAsync(x => x.QuestId == session.DungeonId);
+
+        bool isFirstClear = oldQuestData is null || oldQuestData.PlayCount == 0;
+        bool? oldMissionClear1 = oldQuestData?.IsMissionClear1;
+        bool? oldMissionClear2 = oldQuestData?.IsMissionClear2;
+        bool? oldMissionClear3 = oldQuestData?.IsMissionClear3;
```
SapiensAnatis, line 48: I think DungeonStartController will create the quest in the db, so you could use SingleAsync and ditch the null checking

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
```diff
@@ -92,38 +124,42 @@ public async Task<DragaliaResult> Record(DungeonRecordRecordRequest request)
...
+                                }
                             }
-                        },
+                            : new List<AtgenFirstClearSet>(),
                         take_coin = 1000,
                         take_astral_item_quantity = 300,
-                        missions_clear_set = Enumerable
-                            .Range(1, 3)
+                        missions_clear_set = clearedMissions
+                            .Where((x, index) => (index is > 0 and < 4) && x)
```
SapiensAnatis, line 141: is limiting the index necessary here?

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
```diff
@@ -49,7 +65,23 @@ public async Task<DragaliaResult> Record(DungeonRecordRecordRequest request)
         userData.Exp += 1;
         userData.ManaPoint += 1000;
         userData.Coin += 1000;
-        userData.Crystal += 25;
+
+        bool[] clearedMissions = new bool[5]
+        {
+            isFirstClear,
+            oldMissionClear1 != newQuestData.IsMissionClear1 && newQuestData.IsMissionClear1,
```
SapiensAnatis, line 72: would it be clearer to write 
```suggestion
            newQuestData.IsMissionClear1 && !oldMissionClear1,
```

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
Nightmerp: I think so, otherwise it'll include first clear bonus and all mission cleared bonus into the list as well which might cause problems?

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
Nightmerp: Oh yeah will change those tyty

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
SapiensAnatis: oh right, I misunderstood. Maybe it would be better to keep those variables out of the array in that case? seeing as they're used for different things

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
Nightmerp: Found out that DungeonRecordTest has a bunch of null fields passed into request so ended up keeping it to prevent them from throwing errors

DragaliaAPI/Controllers/Dragalia/DungeonRecordController.cs
Nightmerp: Oh I'll do that then ye, I don't think it should mess with the logic of all missions cleared either since it's impossible for a first clear mission to be the last and only mission necessary to fulfill it

DragaliaAPI.Test/Unit/Controllers/DungeonRecordControllerTest.cs
```diff
@@ -0,0 +1,579 @@
...
+            .Setup(x => x.GetQuests(DeviceAccountId))
+            .Returns(new List<DbQuest>()
+                {
+                    new()
+                    {
+                        DeviceAccountId = DeviceAccountId,
+                        QuestId = questId,
+                        BestClearTime = clearTime
+                    }
+                }.AsQueryable().BuildMock());
```
SapiensAnatis, line 141: You already have a setup for almost this exact same thing in the constructor
```csharp
 this.mockQuestRepository
            .Setup(x => x.GetQuests(DeviceAccountId))
            .Returns(new List<DbQuest>()
                {
                    new() { DeviceAccountId = DeviceAccountId, QuestId = questId }
                }.AsQueryable().BuildMock());
```

It's personal preference more than anything but I like to avoid putting setups in the constructor and keep them in the test methods, so that it's clear reading the method what I'm testing. Unless it's boilerplate stuff and not really relevant (I would say FinishDungeon falls under that umbrella). It's up to you, but it would probably be a good idea to remove the duplicate ones at least.

DragaliaAPI.Test/Unit/Controllers/DungeonRecordControllerTest.cs
```diff
@@ -0,0 +1,579 @@
...
+        this.mockDungeonService.VerifyAll();
+        this.mockUserDataRepository.VerifyAll();
+        this.mockInventoryRepository.VerifyAll();
+        this.mockUpdateDataService.VerifyAll();
+    }
+
+    // Tests that when previous missions have been completed that the right amount of wyrmite is
+    // given as well as their corresponding mission numbers in missions_clear_set
+    [Fact]
+    public async Task VariousPreviousMissionClearStatusesGiveCorrectMissions()
```
SapiensAnatis, line 342: When you're doing multiple test cases and you want to assert similarity on all of them, you can do a [Theory] test. This allows you to pass in parameters to the test function and repeat the same code with different values.

Outside of that, [Fact] unit tests should generally only contain one test case in an arrange -> act -> assert pattern. 

Have a look at [TheoryData](https://hamidmosalla.com/2020/04/05/xunit-part-8-using-theorydata-instead-of-memberdata-and-classdata/) in particular -- you may be able to have a list of values like `GetQuests quest`, `CompleteQuests quest` and `expected missions in the response` and then just repeat the same code of setup repository, call method, check response for each case

DragaliaAPI.Test/Unit/Controllers/DungeonRecordControllerTest.cs
```diff
@@ -0,0 +1,579 @@
...
+                        Name = "Euden",
+                        ViewerId = 1
+                    }
+                }.AsQueryable().BuildMock());
+
+        this.mockInventoryRepository
+            .Setup(
+                x => x.AddMaterials(DeviceAccountId, It.IsAny<List<Materials>>(), It.IsAny<int>())
+            )
+            .Returns(Task.FromResult(0));
```
SapiensAnatis, line 118: I would do `.Returns(Task.CompletedTask)` for plain async Task methods

DragaliaAPI.Test/Unit/Controllers/DungeonRecordControllerTest.cs
```diff
@@ -0,0 +1,579 @@
...
+                        Scene02: "",
+                        AreaName02: "",
+                        Scene03: "",
+                        AreaName03: "",
+                        Scene04: "",
+                        AreaName04: "",
+                        Scene05: "",
+                        AreaName05: "",
+                        Scene06: "",
+                        AreaName06: ""
```
SapiensAnatis, line 73: May be more concise to do 
```suggestion
                    QuestData = MasterAsset.QuestData.Get(100010101)`
```
which is the first campaign quest. This is if you're not particularly interested in asserting what the quest is

DragaliaAPI.Test/Unit/Controllers/DungeonRecordControllerTest.cs
```diff
@@ -0,0 +1,579 @@
...
+        this.dungeonRecordController.SetupMockContext();
+
+        this.mockDungeonService
+            .Setup(x => x.FinishDungeon(dungeonKey))
+            .ReturnsAsync(
+                new DungeonSession()
+                {
+                    Party = new List<PartySettingList>(),
+                    QuestData = new QuestData(
+                        Id: questId,
```
SapiensAnatis, line 54: Could you also add a test to make sure that the party and quest id values appear in the response as expected? Search for quest_party_setting_list and quest_id

SapiensAnatis (changes requested):
Pretty good, especially for your first attempt at unit testing! Just a few minor comments and then we can get this merged in

SapiensAnatis (approved):

PR #70, opened 2023-01-01
Data service rework

### LINKED ISSUE #42: Data services rework
- Should be static classes/methods as opposed to services. Can look at initializing them on startup
- Generic implementation with filename/model type. Then holder class with instances for each needed json file
- Should use exported JSON with all fields and just adapt models to add new fields

SapiensAnatis:
Reworks the various data services to not be distinct types, but instead implementations of a base service which is composed of a KeyedCollection. Full JSON files are used with the models containing the subset of fields that we actually need. This should make it easier to use the data services (they're now static and no injection is required) and it should also make it far easier to extend them, either with new JSON files or with new fields in the model.

Closes #42


DragaliaAPI.Database/Factories/DbPlayerCharaDataFactory.cs
```diff
@@ -53,7 +54,7 @@ internal static DbPlayerCharaData Create(string deviceAccountId, DataAdventurer
             ExAbilityLevel = 1,
             ExAbility2Level = 1,
             IsTemporary = false,
-            IsUnlockEditSkill = data.Availability == "Story",
```
SapiensAnatis, line 56: FIx this before merge

PR #74, opened 2023-01-02
Add /user/withdrawal

SapiensAnatis:
This endpoint is used to delete your savedata in the settings menu. Not exactly sure what the intended server implementation is, but this impl will simply clear the database of all of your data. Hopefully it makes a new DeviceAccount and thus BaaS ID...


SapiensAnatis:
Appears that the client then immediately attempts to call /load/index again without going through /tool/auth, so we'll either need to send an id-token refresh request or make it not totally clear the save

PR #94, opened 2023-01-18
Fixed kelly helper setup

Nightmerp:
Pbrun -> Horus
atk rate go brrrrrrrr


SapiensAnatis:
earth-shattering PR, almost as much as #89

Issue #96, opened 2023-01-18
Wyrmprint functionality


SapiensAnatis:
Wyrmprints should be able to be upgraded and the database updated with these changes.
Should also add ability to buy wyrmprints as part of these changes.

May optionally add wyrmprint sets, but this is lower priority.

Related endpoints:

- [ ]     /ability_crest/buildup_piece
- [ ]     /ability_crest/buildup_plus_count
- [ ]     /ability_crest/get_ability_crest_set_list
- [ ]     /ability_crest/reset_plus_count
- [ ]     /ability_crest/set_ability_crest_set
- [ ]     /ability_crest/set_favorite
- [ ]     /ability_crest_trade/get_list
- [ ]     /ability_crest_trade/trade
- [ ]     /ability_crest/update_ability_crest_set_name

PR #132, opened 2023-02-02
Wyrmprints

Nightmerp:
Trading and upgrading
#96


DragaliaAPI.Database/Repositories/AbilityCrestRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+            x => x.DeviceAccountId == this.playerDetailsService.AccountId
+        );
+
+    public async Task Add(AbilityCrests abilityCrestId)
+    {
+        this.logger.LogDebug("Adding ability crest {print}", abilityCrestId);
+
+        if (await this.FindAsync(abilityCrestId) is not null)
+        {
+            this.logger.LogDebug("Ability crest was already owned.");
```
SapiensAnatis, line 37: Could you make this a warning?

I would think in most cases when we want to add ability crests, we should be doing this check higher up so that we can handle the crest already existing and give a substitute, e.g. for missions and quests you get a token amount of eldwater instead.

DragaliaAPI.Database/Repositories/AbilityCrestRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+            return;
```
SapiensAnatis, line 38: Could you add a test for this case

DragaliaAPI.Database/Repositories/IUserDataRepository.cs
```diff
@@ -21,4 +21,7 @@ public interface IUserDataRepository : IBaseRepository
     Task UpdateCoin(long offset);
     Task<bool> CheckCoin(long quantity);
     Task GiveWyrmite(int quantity);
+    Task UpdateDewpoint(int quantity);
+    Task<bool> CheckDewpoint(int quantity);
+    Task SetDewpoint(int quantity);
```
SapiensAnatis, line 26: In what situation do you need to set the user's eldwater to an absolute value rather than +/- ing it by some value?

DragaliaAPI.Shared.Test/Unit/MasterAssetTest.cs
```diff
@@ -477,4 +477,231 @@ public void StoryData_Type_IsCorrect(int storyId, StoryTypes expectedType)
...
+    [Fact]
+    public void AbilityCrestTrade_Enumerable_ReturnsListOfAbilityCrestTrades()
+    {
+        IEnumerable<AbilityCrestTrade> abilityCrestTrades = MasterAsset
+            .MasterAsset
+            .AbilityCrestTrade
+            .Enumerable;
+
+        abilityCrestTrades.Should().AllBeOfType<AbilityCrestTrade>();
+    }
```
SapiensAnatis, line 507: Not sure what the value of this test is

DragaliaAPI/Services/AbilityCrestService.cs
```diff
@@ -0,0 +1,429 @@
...
+        if (!ValidateAugments(rarityInfo, buildup.plus_count_type, buildup.plus_count))
+        {
+            return ResultCode.AbilityCrestBuildupPlusCountCountError;
+        }
+
+        DbAbilityCrest dbAbilityCrest = await TryFindAsync(abilityCrest.Id);
+        int usedAugments;
+        Dictionary<Materials, int> materialMap;
+
+        if (buildup.plus_count_type == 1)
```
SapiensAnatis, line 227: Can you make these an enum or consts rather than magic numbers

docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
       - ASPNETCORE_URLS=http://+:80
     ports:
       - "80:80"
+      - "433:433"
```
SapiensAnatis, line 19: won't work without updating ASPNETCORE_URLS

I removed this because of development certificate issues, did you add it back or is it just leftover

SapiensAnatis (commented):
This looks great, thank you for all your hard work! Just a few minor comments to address but I think we can release this soon :)

DragaliaAPI.Database/Repositories/IUserDataRepository.cs
Nightmerp: I think this one was just for testing? To set the eldwater value to something that I know won't be enough

docker-compose.yml
Nightmerp: Uhhh I never messed with this so no idea where this came from ;;;

DragaliaAPI.Database/Repositories/AbilityCrestRepository.cs
Nightmerp: Do you mean like LogError instead of LogDebug? Then ye ofc, and also in AbilityCrestService already added the functionality for returning eldwater/rupies if the print is already owned with AddOrRefund ^^

DragaliaAPI.Shared.Test/Unit/MasterAssetTest.cs
Nightmerp: uh. i'm ngl i forgor lmao i'll just remove it

DragaliaAPI.Database/Repositories/AbilityCrestRepository.cs
Nightmerp: Ohhh wait there's a LogWarning too icic

docker-compose.yml
SapiensAnatis: Could you remove it please

DragaliaAPI.Database/Repositories/AbilityCrestRepository.cs
SapiensAnatis: hasn't been done?

DragaliaAPI.Shared.Test/Unit/MasterAssetTest.cs
SapiensAnatis: Yes please

SapiensAnatis (approved):

PR #141, opened 2023-02-10
Use docker containers for integration tests

SapiensAnatis:
- Make DragaliaAPI.Test use a real postgres and redis container using [TestContainers](https://dotnet.testcontainers.org/) locally
- DragaliaAPI.Database.Test still uses an in-memory database
- All integration tests now inherit from TestFixture which provides the `HttpClient` and `ApiContext`, so this should reduce test boilerplate code
- Fix all uses of `DateTime.UtcNow` to `DateTimeOffset.UtcNow`


SapiensAnatis:
Superceded by #202

PR #142, opened 2023-02-13
Halidom implementation

eandrade-dev:
### Original issue
https://github.com/SapiensAnatis/DragaliaAPI/issues/77

### What's here 
* Implemented the following `/fort` endpoints:
   - add_carpenter
   - build_at_once
   - build_cancel
   - build_end
   - build_start
   - levelup_at_once
   - levelup_cancel
   - levelup_end
   - levelup_start
   - move

### What's NOT here
* Two endpoints are still missing or partially implemented
  - `get_multi_income`: Haven't had time to figure this one out properly yet
  - `set_new_fort_plant`: No idea what it is supposed to do. It is called when the "Build" button is tapped, and occasionally crashes the Halidom if not implemented.

### Notes

* `build_cancel` has come erratic graphical behavior - the building completely disappears until the Halidom is reloaded. Works correctly otherwise.
* I could not figure out how to obtain and subtract the user's Diamantium, so those parts currently have placeholders
* Nice to meet you, I'm also on Discord :) Skazord#6728


DragaliaAPI/Controllers/Dragalia/FortController.cs
```diff
@@ -43,13 +70,695 @@ await this.fortRepository.GetBuilds(this.DeviceAccountId).ToListAsync()
...
+                production_st = StubData.ProductionSt,
+                production_df = StubData.ProductionDf,
+                fort_detail = fortDetail,
+                update_data_list = updateDataList,
+            };
+        return this.Ok(data);
+    }
+
+    [HttpPost("build_cancel")]
+    public async Task<DragaliaResult> BuildCancel(FortBuildCancelRequest request)
```
SapiensAnatis, line 221: Could you look at extracting the shared logic between build_* and levelup_* endpoints into a service?

DragaliaAPI.Database/Repositories/FortRepository.cs
```diff
@@ -58,4 +59,64 @@ public async Task<bool> CheckPlantLevel(FortPlants plant, int requiredLevel)
...
+            .FirstAsync();
+        fortDetail.WorkingCarpenterNum = working_carpenter_num;
+        apiContext.Entry(fortDetail).State = EntityState.Modified;
+    }
+
+    public async Task<DbFortBuild> GetBuilding(string accountId, long buildId)
+    {
+        return await apiContext.PlayerFortBuilds
+            .Where(x => x.DeviceAccountId == accountId && x.BuildId == buildId)
+            .FirstAsync();
```
SapiensAnatis, line 105: this will throw an exception if the player tries to request a build ID they don't own -- unlikely to happen but maybe worth adding handling for?

DragaliaAPI/Controllers/Dragalia/FortController.cs
```diff
@@ -43,13 +70,695 @@ await this.fortRepository.GetBuilds(this.DeviceAccountId).ToListAsync()
...
+    private async Task ConsumePlayerMaterials(IQueryable<DbPlayerMaterial> userMaterials, DbPlayerUserData userData, FortPlantDetail plantDetail)
+    {
+        userData.Coin -= plantDetail.Cost;
+
+        // Is there maybe a more efficient way to do this? Seems ugly, but works
+        await ConsumeMaterial(userMaterials, plantDetail.MaterialsId1, plantDetail.MaterialsNum1);
+        await ConsumeMaterial(userMaterials, plantDetail.MaterialsId2, plantDetail.MaterialsNum2);
+        await ConsumeMaterial(userMaterials, plantDetail.MaterialsId3, plantDetail.MaterialsNum3);
+        await ConsumeMaterial(userMaterials, plantDetail.MaterialsId4, plantDetail.MaterialsNum4);
+        await ConsumeMaterial(userMaterials, plantDetail.MaterialsId5, plantDetail.MaterialsNum5);
```
SapiensAnatis, line 699: there is indeed, have a look at https://github.com/SapiensAnatis/DragaliaAPI/blob/develop/DragaliaAPI.Database/Repositories/InventoryRepository.cs#L158

In the weapons case I have built a map of materials in the MasterAsset model: https://github.com/SapiensAnatis/DragaliaAPI/blob/develop/DragaliaAPI.Shared/MasterAsset/Models/WeaponBodyBuildupGroup.cs#L29 and just deduct that

SapiensAnatis (changes requested):
Great first contribution!

Could you also look at adding some tests -- generally for new features I like to have unit tests covering new methods (so repository/service/controllers) and then some functional tests that simulate sample requests (as much as possible anyway). Let me know if you have any questions about writing the tests.

DragaliaAPI.Database/Repositories/FortRepository.cs
```diff
@@ -58,4 +59,64 @@ public async Task<bool> CheckPlantLevel(FortPlants plant, int requiredLevel)
...
+            new DbFortDetail()
+            {
+                DeviceAccountId = accountId,
+                CarpenterNum = 2,
+                MaxCarpenterCount = 5,
+                WorkingCarpenterNum = 0
+            }
+        );
+        await apiContext.SaveChangesAsync();
+        return true;
```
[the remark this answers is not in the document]
eandrade-dev: I think it would be best done that way, yes. As it wasn't done yet, I had to work with what I had for now. It's also the reason why I had to do this:
```
if (!query.Any())
{
    await this.fortRepository.InitFortDetail(this.DeviceAccountId);
    fortDetails = StubData.FortDetail;
}
```
in the Controller, as right now, there's no guarantee that there will be a FortDetail saved for the account on the database. This is merely to ensure that we don't get softlocked when opening the Halidom.

When these changes to the savefile are implemented, this function should be removed entirely.

DragaliaAPI.Database/Entities/DbFortDetail.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace DragaliaAPI.Database.Entities;
+
+[Table("PlayerFortDetail")]
+[Index(nameof(DeviceAccountId))]
+public class DbFortDetail : IDbHasAccountId
+{
+    [Key]
+    [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+    public long FortId { get; set; }
```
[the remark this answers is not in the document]
eandrade-dev: Didn't really think about it at the time, but you are correct. Will change, thank you.

DragaliaAPI/Controllers/Dragalia/FortController.cs
```diff
@@ -43,13 +70,695 @@ await this.fortRepository.GetBuilds(this.DeviceAccountId).ToListAsync()
...
+        int startDateUnix = (int)((DateTimeOffset)startDate).ToUnixTimeSeconds();
+        int endDateUnix = (int)((DateTimeOffset)endDate).ToUnixTimeSeconds();
+        FortBuildStartData data =
+            new()
+            {
+                result = 1,
+                build_id = (ulong)build.BuildId,
+                build_start_date = startDateUnix,
+                build_end_date = endDateUnix,
+                remain_time = endDateUnix - startDateUnix,
```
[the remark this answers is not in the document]
eandrade-dev: `remain_time` expects an `int`, so the alternative using your suggestion would be something like:

`(int)(endDate - startDate).TotalSeconds`

Also worth noting that we would still need to have the `startDateUnix` and `endDateUnix`, as `build_start_date` and `build_end_date` also require `int`s.

Don't think this change would make the code better or easier to read in this instance, unless I am missing something!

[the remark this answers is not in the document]
LukeFZ: there is a custom TimestampToUnixInt formatter for the request/response models (see https://github.com/SapiensAnatis/DragaliaAPI/blob/develop/DragaliaAPI/MessagePack/TimespanToUnixIntFormatter.cs), so by changing the type to a TimeSpan it gets automatically converted to and from an int during serialization.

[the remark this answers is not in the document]
LukeFZ: Note that the same also applies to both startDateUnix and endDateUnix, as those can be changed to DateTimeOffsets which are converted using the custom DateTimeOffsetToUnixInt converter (see https://github.com/SapiensAnatis/DragaliaAPI/blob/develop/DragaliaAPI/MessagePack/DateTimeOffsetToUnixIntFormatter.cs)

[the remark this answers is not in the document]
SapiensAnatis: Oh, sorry I wasn't clear, I also meant you should change the type in the response to `TimeSpan` and it will get serialized as an int (so the game will never know). The types autogenerated and we frequently change them to make them easier to use because a lot of the game's datatypes make no sense (e.g. 0/1 for bools)

DragaliaAPI.Database/Repositories/FortRepository.cs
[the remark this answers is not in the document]
SapiensAnatis: Feel free to change `CreateBase` to do it there, it's not offlimits by any means

DragaliaAPI.Database/Repositories/FortRepository.cs
```diff
@@ -12,7 +7,7 @@
 
 namespace DragaliaAPI.Database.Repositories;
 
-public class FortRepository : IFortRepository
+public class FortRepository : BaseRepository, IFortRepository
```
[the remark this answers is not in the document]
eandrade-dev: Thanks for the tip, makes total sense, changed.

DragaliaAPI/Controllers/Dragalia/FortController.cs
eandrade-dev: Very nifty, managed to adapt something very similar to this case, thanks!

DragaliaAPI/Controllers/Dragalia/FortController.cs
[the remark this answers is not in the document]
eandrade-dev: Ah, that makes a lot of sense now, thanks! I'll make those changes

DragaliaAPI.Database/Repositories/FortRepository.cs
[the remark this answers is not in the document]
eandrade-dev: If a reset or fresh server ever happened, this would no longer be a problem, but I'm unsure if there are plans for that. But to fix the problem for current accounts, the migration would be a good solution.

Maybe this migration and the changes to the SaveFile could be done in a different PR though, as this one is big enough as it is. I would like to keep the band-aid for now, and work on that next, if that's OK with you too.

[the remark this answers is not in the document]
SapiensAnatis: I understand your concerns, but that might make things more complicated. Once this PR is merged in it'll be auto-deployed to the server, so the SQL you write after the fact can't just be a simple INSERT and will have to account for the fact that the data may already exist.

I don't think it is too complicated, something like

```sql
INSERT INTO "PlayerFortDetail" ("DeviceAccountId", "CarpenterNum", "MaxCarpenterCount", "WorkingCarpenterNum")
SELECT "DeviceAccountId", 2, 5, 0 FROM "Players"
```

should work. Additionally, the savefile side would just be moving the code you've already written to a different method.

If you'd like, I am capable of pushing to forks (I think), so once you've finished off your changes I can make this change.

DragaliaAPI.Database/Entities/DbFortDetail.cs
```diff
@@ -0,0 +1,30 @@
...
+    [Required]
+    public int CarpenterNum { get; set; }
+
+    [Column("MaxCarpenterCount")]
+    [Required]
+    public int MaxCarpenterCount { get; set; }
+
+    [Column("WorkingCarpenterNum")]
+    [Required]
+    public int WorkingCarpenterNum { get; set; }
```
SapiensAnatis, line 29: Just had some thoughts about this DB entity in the meantime

1. Is there any particular reason to track MaxCarpenterCount? It should be the same value (5) for all players, right?
2. Would it be better to calculate WorkingCarpenterNum on-demand by querying buildings with a pending `BuildEndDate`, rather than tracking it as an independent column? Concerned about the possibility of it becoming out-of-sync.

Just some thoughts, not requests for change -- let me know what you think.

DragaliaAPI.Database/Repositories/FortRepository.cs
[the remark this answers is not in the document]
eandrade-dev: In that case, I'll be removing this function from the PR altogether. I also don't like leaving loose ends, so it's probably for the best :)

As for the changes to the savefile, if that works for you, it works for me. I'm still getting to grips with the codebase, so if you already know where the code goes, I'm okay with you taking over that part, no problem.

DragaliaAPI.Database/Entities/DbFortDetail.cs
eandrade-dev: For this entity, I based myself on the `FortDetail` class that already existed in `Models\Generated\Components.cs`. Didn't think too critically of it at the time, assumed all fields would be required at some point.

1. The maximum should always be 5, correct. I would be in favor of removing the field from the entity and storing it as a constant wherever it makes sense, I believe it would be the actual Controller in this case.
2. The possibility of desync is real, I ran into it multiple times when implementing these endpoints. In theory, if everything works correctly, it _should_ never happen. I'm not opposed to querying all the builds, but I'm concerned about the performance of that. Wouldn't it be too expensive to run such a query every time? Do note that we have to feed this info on the `/fort/get_data` endpoint, which runs every time someone opens the Halidom menu and the Halidom itself, not just when a build is started/finished.

SapiensAnatis: Constant in the controller is fine — the config setting is just a wild idea, and I’m not even sure changing it would work; I vaguely remember seeing defined levels in the client data, presumably it has to know how much the upgrade will cost without asking the server. 

For the active count, I’m sure your code works fine, it’s more a concern if changes are made to this later —seems like the sort of thing that could be easily missed. However do you know if this entity appears in update_data_list? If it does, having it as calculated rather than a db entity may make things a bit tricky.

eandrade-dev: Currently, I'm only using the MaxCarpenterNum on the Controller, which is why I suggested that. If a Service makes more sense, I'm fine with moving it, though.

Regarding active count: It does not seem to be included in `update_data_list`. The client gets the info from the `fort_detail`, which is a field part of the response in most, if not all of these endpoints. So I think we would be safe on that end.

I can make those changes so that we can test the performance hit, shouldn't be too much work. If it's bad, we can always go back on them, no big deal.

eandrade-dev: Just pushed these changes and a few more requested in the PR.

About the constant, there is currently no service for the Fort implementation, and I'm unsure if there should be one or not. For now, I've moved the constant to the Repository where it probably makes more sense than the controller.

SapiensAnatis: Yeah I think a separate fort service should be good. The usual structure I try to go with is:

- controller parses the request, calls the service, and formats the response
- service performs all the logic behind the response (checking mats, making repository calls)
- repository provides methods to update the database

DragaliaAPI.Database/Repositories/FortRepository.cs
```diff
@@ -38,6 +35,11 @@ ILogger<FortRepository> logger
             x => x.DeviceAccountId == this.playerDetailsService.AccountId
         );
 
+    public IQueryable<DbFortDetail> Details =>
+        this.apiContext.PlayerFortDetails.Where(
+            x => x.DeviceAccountId == this.playerDetailsService.AccountId
```
SapiensAnatis, line 40: When searching for a single entity based on its primary key, it is usually better to use FindAsync, as that will avoid a trip to the DB if it's already being tracked

DragaliaAPI/AutoMapper/Profiles/FortMapProfile.cs
```diff
@@ -10,6 +10,7 @@ public FortMapProfile()
     {
         this.CreateMap<DbFortBuild, FortPlantList>();
         this.CreateMap<DbFortBuild, BuildList>();
+        this.CreateMap<DbFortDetail, FortDetail>();
```
SapiensAnatis, line 13: Instead of providing an operator in the generated class, you could do

```suggestion
        this.CreateMap<DbFortDetail, FortDetail>()
        .ForMember(x => x.working_carpenter_num, opts => opts.MapFrom(src => 5));
```

SapiensAnatis:
It would be good to pull out the fort logic into a service, I think, and then could you add unit tests for that?

You could also add unit tests for the repository but I don't want to ask too much!

Once you've done that and addressed the few other nitpicks I had, I will be happy to merge this in :)

DragaliaAPI/AutoMapper/Profiles/FortMapProfile.cs
SapiensAnatis: Actually I've just seen the test pipeline -- you will have to do this to stop automapper complaining :innocent:

DragaliaAPI.Database/Entities/DbFortDetail.cs
eandrade-dev: Should be a bit of a refactor work, but I think it makes more sense than the current implementation, I'll get it done

SapiensAnatis (approved):

PR #153, opened 2023-02-26
Fixes for Dragon LimitBreak and non-levelup gifts

SteelGram:


DragaliaAPI/Services/DragonService.cs
```diff
@@ -465,12 +470,14 @@ await RollPuppyThankYouRewards(request.dragon_gift_id_list, deviceAccountId)
                     return_gift_list = rewards.Where(x => x.Item1 == gift).First().Item2,
                     dragon_gift_id = gift,
                     is_favorite =
-                        dragonData.FavoriteType != null
-                        && DragonConstants.rotatingGifts[(int)dragonData.FavoriteType] == gift,
-                    reward_reliability_list = levelGifts.Where(x => x.Item1 == gift).First().Item2
+                        DragonConstants.rotatingGifts[(int)dragonData.FavoriteType] == gift,
+                    reward_reliability_list = levelGifts.Any(x => x.Item1 == gift)
+                        ? levelGifts.Where(x => x.Item1 == gift).First().Item2
+                        : null
```
SapiensAnatis, line 476:
```suggestion
                    reward_reliability_list = levelGifts.FirstOrDefault(x => x.Item1 == gift)?.Item2
```

DragaliaAPI/Services/DragonService.cs
```diff
@@ -566,10 +573,15 @@ await DoDragonGetContactData(new(), deviceAccountId)
...
+        logger.LogDebug(
+            "Creating response from rewards {@rewards} and levelGifts: {@levelGifts}",
+            rewards,
+            levelGifts
+        );
         return new DragonSendGiftMultipleData()
         {
             is_favorite = true,
-            reward_reliability_list = levelGifts.First().Item2,
+            reward_reliability_list = levelGifts.Any() ? levelGifts.First().Item2 : null,
```
SapiensAnatis, line 584:
```suggestion
            reward_reliability_list = levelGifts.FirstOrDefault()?.Item2,
```

SapiensAnatis (changes requested):
Can you add a test for the case where reliability is increased but the level doesn't increased, checking that the gift list is null?

p.s. is that the correct behaviour or should it be empty? Have you tested these changes on an actual client

SapiensAnatis (approved):

PR #158, opened 2023-03-14
Photon test

SapiensAnatis:
- Adds the photon plugin to the monorepo
   - This can be debugged locally, it expects a PhotonServer/ folder in the solution directory
- Sends `photon.dawnshard.co.uk` as the multi server host from `/load/index` to point at a test instance of photon


SapiensAnatis:
Shut up codecov lol

PR #161, opened 2023-03-20
Fixed missing initial charastory on new chara add

SteelGram:


DragaliaAPI.Database/Repositories/UnitRepository.cs
```diff
@@ -103,6 +112,38 @@ IEnumerable<Charas> idList
...
+                            DeviceAccountId = deviceAccountId,
+                            StoryType = StoryTypes.Chara,
+                            StoryId = story.storyIds[0],
+                            State = 0
+                        }
+                    );
+                }
+                else
+                {
+                    logger.LogWarning(
```
SapiensAnatis, line 140: Can you make this information as it doesn't necessarily indicate an error or unexpected behaviour

Or maybe log `adding chara X`, then `adding stories Y and Z`, then just not logging the latter in this case

SapiensAnatis:
Thanks for supplying a fix for this, really appreciate the quick work

DragaliaAPI.Database/Repositories/UnitRepository.cs
SteelGram: Addressed in 1b57798

PR #164, opened 2023-04-23
Fix build max level upgrade

eandrade-dev:
This merge request reverts the previous change where leveling of buildings was done when leveling finished, which conflicted with the client app and caused exceptions

Leveling is now done when building/upgrading starts, in line with how the app handles this as well

Also fixed the relevant unit tests


SapiensAnatis (approved):
Thank you for looking into this :)

PR #203, opened 2023-05-08
Fix for levelling up level 1 buildings

SapiensAnatis:
This was causing a primary key exception due to FortService attempting to add a build that already existed:

image


BatemanBatman:
hey guys im a new here can you explain me a little bit!!!

PR #217, opened 2023-05-29
implement fort/get_multi_income

sockperson:
Implement collecting resources from Rupie Mines and Dragontrees in the Halidom

- should I move a lot of this logic into DragaliaAPI/Services/FortService?
- not too sure how to do tests for this


SapiensAnatis:
Work incorporated as part of #270

PR #230, opened 2023-06-12
Implement tutorial item grants and various bugfixes

LukeFZ:
Implements a TutorialService that handles tutorial states and flags, alongside managing the one-time-drops for tutorials such as Halidom or Wyrmprints.

Also implements a (dev-only!) savefile/export Endpoint which enables the export of savedata, along with fixing a few bugs in the JSON time converters.

Reworks the logic for selecting Halidom building updates so that it now supports level 0 decorations which, in addition to being level 0, also take no time to build.

Makes CharaController now check if the upgraded character has stories before adding them so that Euden can now be upgraded.

As for new features, implements the item summon odds endpoint backed by a config specifying the current item rates.


DragaliaAPI/Services/Game/TutorialService.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        switch (storyId)
+        {
+            case 1000106:
+                await this.userDataRepository.UpdateDewpoint(100);
+                await this.inventoryRepository.UpdateQuantity(Materials.HolyWater, 10);
+                if (await this.abilityCrestRepository.FindAsync(AbilityCrests.ManaFount) == null)
+                    await this.abilityCrestRepository.Add(AbilityCrests.ManaFount);
+                logger.LogDebug("Added materials for the wyrmprint tutorial");
+                break;
```
SapiensAnatis, line 64: Not the biggest fan of this in terms of readability becaus I can see it getting huge. But I guess at some point you need to branch based on the cleared story... 

Do you think it would look nicer if we used constants for the ids and kept the switch small by adding methods?

```suggestion
            case StoryIds.Chapter1_1:
                await this.SetupWyrmprintTutorial()
                break;
```

SapiensAnatis (approved):

PR #236, opened 2023-06-14
Replace all methods accepting DeviceAccountId

LukeFZ:
Replacement is the new IPlayerIdentityService, which handles impersonation and provides the AccountId to everything that needs it.

PR #238, opened 2023-06-15
Make save importing not as slow

LukeFZ:
Makes save importing faster due to the use of parallelization and changing the UnitStory mapper to no longer do a MasterData lookup and enum value check to determine the type (Chara or Dragon).

Depends on #236


DragaliaAPI/Services/Game/SavefileService.cs
```diff
@@ -117,13 +117,24 @@ public async Task Import(LoadIndexData savefile)
             this.playerIdentityService.AccountId
         );
 
+        this.apiContext.ChangeTracker.AutoDetectChangesEnabled = false;
```
SapiensAnatis, line 120: How come you turned this off?

LukeFZ: Also increases performance (gained around ~1sec in testing iirc) and nothing in here needs change tracking

SapiensAnatis (approved):

PR #254, opened 2023-06-18
Implement endeavor infrastructure

LukeFZ:
(Missions == Endeavors)
Begins work on fully implementing endeavors.
- MissionService is responsible for managing missions and their rewards, along with both drill (royal regimen) and main story grouped missions
- MissionProgressionService is responsible for handling events that progress missions as well as the actual progression itself
- MissionInitialProgressionService is responsible for deriving the initial progress missions start at, for missions that might've been completed prior to obtaining it in the game

This does not fully implement missions, as it only includes mission progression info for the main story progression-locking missions, as well as some of the royal regimen stage 1 missions. Alongside that, there a few events missing to properly progress some unique mission types, and no mechanic to grant daily/normal/special endeavors.

TODO: Tests, Merge conflicts related to db migrations


DragaliaAPI/Features/Missions/MissionProgressionService.cs
```diff
@@ -0,0 +1,164 @@
...
+    public async Task ProcessMissionEvents()
+    {
+        if (this.eventQueue.Count == 0)
+            return;
+
+        List<DbPlayerMission>? missionList = null;
+
+        while (this.eventQueue.TryDequeue(out Event evt))
+        {
+            if (
```
SapiensAnatis, line 86: Would it be worth debug logging every event we process?

SapiensAnatis (approved):

PR #260, opened 2023-06-20
Implement item summoning

LukeFZ:
- Missing tests
- Currently made the shop db addition a new savefile version, could change this back to award on import if wanted


DragaliaAPI.Test/Features/Shop/PaymentServiceTest.cs
```diff
@@ -0,0 +1,170 @@
...
+                        DeviceAccountId = IdentityTestUtils.DeviceAccountId,
+                        BuildTimePoint = type == PaymentTypes.HalidomHustleHammer ? total : 0,
+                        Coin = type == PaymentTypes.Coin ? total : 0,
+                        Crystal = type == PaymentTypes.Wyrmite ? total : 0,
+                        ManaPoint = type == PaymentTypes.ManaPoint ? total : 0,
+                    }
+                }
+                    .AsQueryable()
+                    .BuildMock()
+            );
```
SapiensAnatis, line 54: I added a helper to simplify this a bit: https://github.com/SapiensAnatis/Dawnshard/blob/develop/DragaliaAPI.Test/UnitTestUtils.cs#L121

You don't have to change this but just FYI

DragaliaAPI.Test/Features/Shop/PaymentServiceTest.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        await this.paymentService.ProcessPayment(type, new PaymentTarget(total, cost));
+
+        this.mockUserDataRepository.VerifyAll();
```
SapiensAnatis, line 58: If you assigned the UserData here I think it would get modified by reference and you could check the subtraction

SapiensAnatis (approved):

PR #264, opened 2023-06-21
Ability Crest Sets

Nightmerp:


DragaliaAPI.Database/ApiContext.cs
```diff
@@ -67,6 +67,10 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
         modelBuilder.Entity<DbQuest>().HasKey(e => new { e.DeviceAccountId, e.QuestId });
 
+        modelBuilder
+            .Entity<DbAbilityCrestSet>()
+            .HasKey(key => new { key.DeviceAccountId, key.AbilityCrestSetNo });
+
```
LukeFZ, line 73: you can use [PrimaryKey(nameof(DeviceAccountId), nameof(AbilityCrestSetNo))] on the db class instead

DragaliaAPI/Controllers/Dragalia/AbilityCrestController.cs
```diff
@@ -152,4 +156,88 @@ public async Task<DragaliaResult> ResetPlusCount(AbilityCrestResetPlusCountReque
...
+        int index = 0;
+
+        IEnumerable<AbilityCrestSetList> abilityCrestSetList = Enumerable
+            .Range(1, 54)
+            .Select(
+                x =>
+                    index < dbAbilityCrestSets.Count()
+                    && dbAbilityCrestSets[index].AbilityCrestSetNo == x
+                        ? dbAbilityCrestSets[index++]
+                        : new DbAbilityCrestSet(DeviceAccountId, x)
```
LukeFZ, line 179: This doesn't respect user impersonation, though the value of DeviceAccountId is also never used so its fine - however I'd replace it with an empty string as a precaution if we do decide to completely remove the DeviceAccountId property in the future

DragaliaAPI/Controllers/Dragalia/AbilityCrestController.cs
```diff
@@ -152,4 +156,88 @@ public async Task<DragaliaResult> ResetPlusCount(AbilityCrestResetPlusCountReque
...
+    )
+    {
+        if (request.ability_crest_set_no is <= 0 or > 54)
+        {
+            this.logger.LogError("Invalid ability crest no", request.ability_crest_set_no);
+            return this.Code(ResultCode.CommonInvalidArgument);
+        }
+
+        DbAbilityCrestSet newAbilityCrestSet = mapper.Map<DbAbilityCrestSet>(request);
+        newAbilityCrestSet.DeviceAccountId = DeviceAccountId;
```
LukeFZ, line 205: This also doesn't respect user impersonation and in this case it does matter, and as such I'd recommend making IAbilityCrestRepository::AddOrUpdateSet set the DeviceAccountId of the passed-in db record so it's always assigned to the correct user.

DragaliaAPI/Controllers/Dragalia/AbilityCrestController.cs
```diff
@@ -152,4 +156,88 @@ public async Task<DragaliaResult> ResetPlusCount(AbilityCrestResetPlusCountReque
...
+        DbAbilityCrestSet? dbAbilityCrestSet = await abilityCrestRepository.FindSetAsync(
+            request.ability_crest_set_no
+        );
+
+        if (dbAbilityCrestSet is null)
+        {
+            await abilityCrestRepository.AddOrUpdateSet(
+                new DbAbilityCrestSet()
+                {
+                    DeviceAccountId = DeviceAccountId,
```
LukeFZ, line 227: Same here as in the method above

LukeFZ (changes requested):

LukeFZ (approved):

LukeFZ (approved):

Issue #55, opened 2022-12-25
Daily reset softlock


SapiensAnatis:
**Describe the bug**
If a device is left to idle overnight, and opened after 'reset', a message will appear stating 'a new day has dawned' and kick the player to the home screen (not the title screen). This is despite the server not giving any indication of a daily reset and not even having any concept of one -- perhaps it's based on the client's stored last login time.

Upon returning to the home screen, the user is UI locked with the following screen, softlocked with a glitched endeavour list

![Screenshot_20221225-181642_Dragalia](https://user-images.githubusercontent.com/5047192/209478548-225c0320-48b5-4cfb-8369-da9e0273cf06.jpg)

**On branch**
develop / deployed server 25/12/2022

**To Reproduce**
1. Log into the server
2. Leave the device idling past reset time in your locale
3. Get kicked to title screen

**Expected behavior**
At the very least it shouldn't softlock, but it looks like it's expecting an item in an UpdateDataList somewhere, so maybe we need to fix this by implementing login bonuses.

**Additional context**
ADB logcat:

```
12-25 18:16:39.943 24600 24645 E Unity   : NullReferenceException: Object reference not set to an instance of an object.
12-25 18:16:39.943 24600 24645 E Unity   :   at Gluon.DataManager.UpdateByUpdateDataList (Gluon.Http.UpdateDataList updateDataList) [0x00000] in <00000000000000000000000000000000>:0
12-25 18:16:39.943 24600 24645 E Unity   :   at Gluon.DataManager.UpdateByResponse (Gluon.Http.MypageInfoResponse res) [0x00000] in <00000000000000000000000000000000>:0
12-25 18:16:39.943 24600 24645 E Unity   :   at Gluon.MyPageScene.OnMyPageInfoSuccess (Gluon.Http.MypageInfoResponse response) [0x00000] in <00000000000000000000000000000000>:0
12-25 18:16:39.943 24600 24645 E Unity   :   at System.Action`1[T].Invoke (T obj) [0x00000] in <00000000000000000000000000000000>:0
12-25 18:16:39.943 24600 24645 E Unity   :   at System.Action`1[T].Invoke (T obj) [0x00000] in <00000000000000000000000000000000>:0
12-25 18:16:39.943 24600 24645 E Unity   :   at Gluon.Http.MypageInfoTask.Deserialize (System.Byte[] body) [0x00000] in <00000000000000000000000000000000>:0
[...]

PR #282, opened 2023-06-25
Implement daily bonus

### LINKED ISSUE #211: Implement daily login bonuses
The server does not currently implement daily login bonuses.

These are sent from [/login/index](https://dragalia-api-docs.readthedocs.io/en/latest/dragalia/login_index.html)

There is already some logic to handle 'daily resets' for dragon gifts, although this may need to be reworked.

It is fine for the gifts to be given directly to users rather than going in the gift box, until this functionality is added.

LukeFZ:
Closes #211 #55


DragaliaAPI/Features/Login/LoginController.cs
```diff
@@ -15,27 +16,37 @@ namespace DragaliaAPI.Features.Login;
 [Consumes("application/octet-stream")]
 [Produces("application/octet-stream")]
 [ApiController]
+[BypassDailyReset]
```
SapiensAnatis, line 19: Should this be applied to any other controllers? Anything before you get to /login/index might throw you into an infinite loop, right?

DragaliaAPI/Middleware/DailyResetMiddleware.cs
```diff
@@ -0,0 +1,62 @@
...
+        )
+        {
+            string? sessionId = value.FirstOrDefault();
+            if (sessionId != null)
+            {
+                using IServiceScope scope = serviceProvider.CreateScope();
+
+                ISessionService sessionService =
+                    scope.ServiceProvider.GetRequiredService<ISessionService>();
+                IResetHelper resetHelper = scope.ServiceProvider.GetRequiredService<IResetHelper>();
```
SapiensAnatis, line 38: looked into it and I think you can inject these in the parameters of InvokeAsync? worth a go

DragaliaAPI/Features/Login/LoginController.cs
LukeFZ: DailyResetMiddleware additionally uses the AllowAnonymous attribute for filtering out endpoints that should always be available, meaning that before /tool/auth is called (which refreshes the login time in the session so unblocks any blocked endpoint) all endpoints are accessible

DragaliaAPI/Services/Game/SessionService.cs
```diff
@@ -139,15 +144,20 @@ public async Task<string> ActivateSession(string idToken)
...
+        string idToken,
+        string accountId,
+        long viewerId,
+        DateTimeOffset loginTime
+    )
     {
         // Check for existing session
         Session? existingSession = await this.TryLoadSession(Schema.Session_IdToken(idToken));
         if (existingSession is not null)
             return existingSession.SessionId;
```
SapiensAnatis, line 157: If there's an existing session it may not refresh the login time? 🤔 

Maybe best just to ignore any existing sessions and always issue a new one. That's what the official game did iirc

SapiensAnatis (approved):

PR #297, opened 2023-07-02
Fort AddToStorage fixes + csharpier upgrade

LukeFZ:


DragaliaAPI/Features/Fort/FortRepository.cs
```diff
@@ -203,21 +197,38 @@ public async Task AddBuild(DbFortBuild build)
         await apiContext.PlayerFortBuilds.AddAsync(build);
     }
 
-    public async Task AddToStorage(FortPlants plant, int level)
+    public async Task AddToStorage(
+        FortPlants plant,
+        int? level = null,
+        int quantity = 1,
```
SapiensAnatis, line 203: would it be better to swap these to put the optional args last?

SapiensAnatis (approved):

SapiensAnatis (approved):

PR #317, opened 2023-07-23
Add Player leveling, item storage/consumption, stamina

LukeFZ:


DragaliaAPI/Features/Item/ItemService.cs
```diff
@@ -0,0 +1,62 @@
...
+        logger.LogDebug("Processing items {@useItems}", items);
+
+        UseItemEffect effect = UseItemEffect.None;
+        int totalQuantity = 0;
+
+        foreach (AtgenUseItemList item in items)
+        {
+            UseItemData data = MasterAsset.UseItem[item.item_id];
+            if (effect == UseItemEffect.None)
+                effect = data.ItemEffect;
```
SapiensAnatis, line 36: are we expecting to only get items of the same effect?

DragaliaAPI/Features/Player/UserService.cs
```diff
@@ -0,0 +1,215 @@
...
+            if (!MasterAsset.UserLevel.TryGetValue(data.Level + 1, out UserLevel? next))
+                break;
+
+            if (current.TotalExp + current.NecessaryExp > data.Exp)
+                break;
+
+            data.Level++;
+            totalReward += WyrmiteLevelUpReward;
+
+            await AddStamina(StaminaType.Single, next.StaminaSingle);
```
SapiensAnatis, line 208: does this need to add wings too?

DragaliaAPI/Features/Item/ItemService.cs
LukeFZ: item/use_recovery_stamina doesn't return an array of AtgenRecoverData so the game prevents you from mixing them

DragaliaAPI/Features/Dungeon/DungeonRecordController.cs
```diff
@@ -72,6 +74,33 @@ PlayRecord playRecord
 
         bool isFirstClear = oldQuestData is null || oldQuestData?.PlayCount == 0;
 
+        if (
+            !isFirstClear && (playRecord.is_clear == 1 || session.QuestData.IsPayForceStaminaSingle)
```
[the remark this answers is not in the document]
LukeFZ: Have not verified this but I thought the wave quests did, the ones where stamina is also consumed even if you fail

DragaliaAPI/Features/Player/UserService.cs
LukeFZ: yes

DragaliaAPI/Features/Dungeon/DungeonRecordController.cs
```diff
@@ -72,6 +74,33 @@ PlayRecord playRecord
...
+            StaminaType type = StaminaType.None;
+            int amount = 0;
+
+            if (isMulti)
+            {
+                if (session.IsHost)
+                {
+                    type = StaminaType.Multi;
+                    amount = session.QuestData.PayStaminaMulti;
+                }
```
SapiensAnatis, line 92: Other players also have getherwings deducted

DragaliaAPI/Features/Dungeon/DungeonRecordController.cs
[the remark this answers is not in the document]
SapiensAnatis: Looks like is_clear is set as 1 for incomplete facility clears, I suggest we remove it from this condition if we don't know what it does

SapiensAnatis (approved):

PR #323, opened 2023-07-25
Various event compendium fixes

LukeFZ:


DragaliaAPI/Features/Dungeon/DungeonStartService.cs
```diff
@@ -163,27 +163,25 @@ List<DbDetailedPartyUnit> detailedPartyUnits
...
+            unit.weapon_skin_data ??= new GameWeaponSkin();
+            unit.weapon_body_data ??= new GameWeaponBody();
+            unit.crest_slot_type_1_crest_list ??= Enumerable.Empty<GameAbilityCrest>();
+            unit.crest_slot_type_2_crest_list ??= Enumerable.Empty<GameAbilityCrest>();
+            unit.crest_slot_type_3_crest_list ??= Enumerable.Empty<GameAbilityCrest>();
+            unit.talisman_data ??= new TalismanList();
+            unit.edit_skill_1_chara_data ??= new EditSkillCharaData();
+            unit.edit_skill_2_chara_data ??= new EditSkillCharaData();
+            unit.game_weapon_passive_ability_list ??= Enumerable.Empty<WeaponPassiveAbilityList>();
+        }
```
SapiensAnatis, line 183: Is this needed as well as the default values for the properties? If we can get away with just default values would be a bit neater

SapiensAnatis (approved):

PR #325, opened 2023-07-27
Add tracking for login bonuses

### LINKED ISSUE #299: Login bonus should advance per days logged in, not elapsed
Currently login bonuses will give you the Day 3 bonus if it is 3 days since the start of the bonus (with a modulus in the case of looping bonuses). 

On the official servers, you would only get Day 3 if you logged in 3 times, so you could get Day 3 on e.g. Day 5.

We need to track the number of logins against each bonus, possibly using a new database table or Redis key, and award the next one in the sequence.

SapiensAnatis:
- Add table to track login bonuses, so that players receive the next one they're eligible for rather than the current day.

Closes #299


DragaliaAPI/Features/Login/LoginBonusService.cs
```diff
@@ -36,17 +33,28 @@ public async Task<IEnumerable<AtgenLoginBonusList>> RewardLoginBonus()
...
+            {
+                logger.LogDebug(
+                    "Player has already completed login bonus {@bonus}, skipping...",
+                    dbBonus
+                );
+                continue;
+            }
 
-            dayId += 1;
+            dbBonus.CurrentDay += 1;
```
LukeFZ, line 46: this addition cannot preceed the modulo operation as it is meant to prevent dayId == 0 which can still happen here

LukeFZ (changes requested):

DragaliaAPI/Features/Login/LoginBonusService.cs
SapiensAnatis: moved

LukeFZ (approved):

PR #359, opened 2023-08-02
Make dmode max out skill and ability levels

LukeFZ:
Not entirely sure about this logic, someone double check this

### DIFF
--- a/DragaliaAPI/Features/DmodeDungeon/DmodeDungeonService.cs
+++ b/DragaliaAPI/Features/DmodeDungeon/DmodeDungeonService.cs
@@ -58,18 +58,20 @@ public class DmodeDungeonService(
             x => x.CharaId == charaId
         );
 
+        CharaData charaData = MasterAsset.CharaData[charaId];
+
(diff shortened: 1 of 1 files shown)


DragaliaAPI/Features/DmodeDungeon/DmodeDungeonService.cs
```diff
@@ -58,18 +58,20 @@ await dmodeRepository.Charas.SingleOrDefaultAsync(x => x.CharaId == charaId)
...
-                chara.ComboBuildupCount
+                charaData.HasManaSpiral ? 4 : 3,
+                charaData.HasManaSpiral ? 3 : 2,
+                charaData.MaxAbility1Level,
+                charaData.MaxAbility2Level,
+                charaData.MaxAbility3Level,
+                5,
+                5,
+                2,
+                0
```
SapiensAnatis, line 74: ComboBuildupCount = upgraded standard attack damage, this should be 1 for spiral units

SapiensAnatis (approved):

PR #371, opened 2023-08-05
Fix co-op retry bugs

### LINKED ISSUE #361: Some players in co-op start before others
Some players can move in co-op before the host loads in, either because they are on 'Waiting for other players' or a loading screen. The game is in a glitchy state before this happens. We may need to wait for some event from the host to start the game, possibly by firing another event

### LINKED ISSUE #362: Player backline visible in co-op instead of other players
Seemingly at random sometimes the player's backline is loaded instead of other players in the co-op room. You can control these units and they often die quickly with a typical co-op team layout.

Possibly related to MemberCountTable sent in the PartyData event.

SapiensAnatis:
Closes #361 and closes #362 

Also move SetGameState call above `base.OnLeave` call to prevent `Plugin GluonPlugin tries to set game state after call to 'Continue'` error.

---
 
The root cause of both issues was the `StartQuest` property not being cleared between quest attempts, causing an early start and the proper data to not be received. This is a plugin-only property we use to decide when all players are loaded and to fire the StartQuest event.

However, when clearing `StartQuest` on a quest retry, the host appears to progress through the `GoToIngameState` values _much_ faster than other players. This leads to other players experiencing an infinite load, as they never receive the required events -- which probably lead to #362 since by not clearing `StartQuest` they would be forced to load into the game without this info.

~~To solve this, the events are now broadcast after a 3-second delay. This is a bit of a hack, maybe it would be better to find some way to only send the `GoToIngameState` progression events on the last player reaching a particular value, but I couldn't find a nice way to do this.~~

Changed approach: uses an instance variable now to track the min GoToIngameState value. No need to use `Interlocked` methods as plugin callbacks are thread-safe.

### DIFF
--- a/DragaliaAPI.Photon.Plugin/Constants/ActorPropertyKeys.cs
+++ b/DragaliaAPI.Photon.Plugin/Constants/ActorPropertyKeys.cs
@@ -31,7 +31,5 @@ namespace DragaliaAPI.Photon.Plugin.Constants
         public const string HeroParamCount = "PLUGIN_HeroParamCount";
 
         public const string MemberCount = "PLUGIN_MemberCount";
-
-        public const string FailQuest = "PLUGIN_FailQuest";
     }
 }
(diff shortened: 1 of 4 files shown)


LukeFZ (approved):
hack:tm: (wanted to show that I'm fine with the hack even if its still a draft)

PR #392, opened 2023-08-13
Add skip ticket support

SapiensAnatis:


SapiensAnatis:
Think this has a bit of an issue: doing the daily bonus quick skip only gives bonuses for one of the quests. Could be related to how the bonuses work or how the skip tickets work.

LukeFZ (approved):

PR #395, opened 2023-08-15
Refactor CharaController and add chara quest experience

LukeFZ:


DragaliaAPI/Features/Dungeon/Record/DungeonRecordService.cs
```diff
@@ -113,13 +115,13 @@ private void ProcessClearTime(IngameResultData resultData, float clearTime, DbQu
...
         growRecord.mana_bonus_factor = 1;
-        growRecord.chara_grow_record = session.Party.Select(
-            x => new AtgenCharaGrowRecord() { chara_id = x.chara_id, take_exp = 1 }
-        );
+        // growRecord.chara_grow_record = session.Party.Select(
+        //    x => new AtgenCharaGrowRecord() { chara_id = x.chara_id, take_exp = 1 }
+        // );
 
         return Task.CompletedTask;
     }
```
SapiensAnatis, line 127: Think this can be removed?

DragaliaAPI/Features/Chara/CharaController.cs
```diff
@@ -693,102 +684,88 @@ await storyRepository.Stories
...
+            if (isOmnicite)
+                continue;
+
             bool isOnlyUsingGrowMaterial =
                 charaData.GrowMaterialOnlyStartDate <= time
                 && charaData.GrowMaterialOnlyEndDate >= time;
 
-            // they smoked some shit
-            switch (upgradeMaterialType)
+            if (isOnlyUsingGrowMaterial || useGrowMaterial)
```
SapiensAnatis, line 695: What causes request.is_use_grow_material to be set? I would assume that also encompasses the logic about whether it's active?

LukeFZ: the "only" period is used for events where you can only upgrade the chara with convictions, and is_use_grow_material should be used when upgrading the chara with convictions after that point to my knowledge

DragaliaAPI/Features/Dungeon/Record/DungeonRecordService.cs
LukeFZ: alright, I left it in since take_player_exp was also implemented previously but left in here

SapiensAnatis (approved):

SapiensAnatis (approved):

PR #407, opened 2023-09-03
Quest bonus rework

SapiensAnatis:


SapiensAnatis:
Closing to do #408 and ToTM fix separately

PR #410, opened 2023-09-03
Use Respawn to clear up databases in tests

SapiensAnatis:
[Respawn](https://github.com/jbogard/Respawn) should be faster than `EnsureCreated()`/`EnsureDeleted()`. 

Some tests have been updated due to relying on specific values of identity columns (e.g. viewer id, key id) which now are dynamic, as tables are not deleted between test classes

### DIFF
--- a/DragaliaAPI.Integration.Test/Features/GraphQL/GraphQlTest.cs
+++ b/DragaliaAPI.Integration.Test/Features/GraphQL/GraphQlTest.cs
@@ -15,7 +15,7 @@ public class GraphQlTest : GraphQlTestFixture
 {
     private const string Endpoint = "savefile/graphql";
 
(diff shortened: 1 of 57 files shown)


DragaliaAPI.Integration.Test/Features/GraphQL/GraphQlTest.cs
```diff
@@ -40,9 +40,9 @@ public async Task Query_PlayerCharacters_ReturnsExpectedResult()
         GraphQLResponse<Response> response = await this.GraphQlHttpClient.SendQueryAsync<Response>(
             new GraphQLRequest
             {
-                Query = """
+                Query = $$"""
                 query {
-                    player(viewerId: 1) {
+                    player(viewerId: {{ViewerId}}) {
```
SapiensAnatis, line 45: Yes I know graphql has variables but it was refusing to work and kept throwing "error occured" on mutations that were fine in Postman

PR #426, opened 2023-09-30
Send decrypt error instead of session token refresh for expiry

SapiensAnatis:


SapiensAnatis:
This doesn't cause the client to go back to baas unfortunately -- maybe 'decrypt error' is coneshell related. #437 should hopefully resolve the issue

PR #429, opened 2023-10-01
Update TalismanValidator.cs

Arceus922:
The portrait wyrmprints without abilities can actually reach up to 100 bonus attack and hp! It should be wrong to exclude such wyrmprints even if they arent optimal! Though the correct hack check should be 0 abilities -> 0-100, 1 ability -> 0-50 , 2 abilities 0 !


SapiensAnatis:
According to https://dragalialost.wiki/w/Enter_the_Kaleidoscape#Portrait_wyrmprints the correct validation is

image

Arceus922:
yeah those might be the in game the values but the true bonus values are actually different!
image

SapiensAnatis:
Ah ok, so they have base values of 20/10 and then the additional values are bonuses to those stats based on the number of abilities.

Thanks for the PR. I'm happy to accept without checking the # of abilities since I don't think 100 strength will make much of a difference in time attack. You will need to update PartyInfoValidatorTest -> InvalidTalismanStats_ReturnsFail for these changes.

SapiensAnatis (approved):

PR #430, opened 2023-10-01
Update PartyInfoValidatorTest.cs

Arceus922:
#429 additional change


SapiensAnatis:
Should be an additional commit in #429 -- does the web editor not let you do that?

Arceus922:
should be, am not quite fammiliar with github tbh

SapiensAnatis:
I've cherrypicked it onto that branch for you

PR #445, opened 2023-10-09
Revert "Disable change tracking for big queries"

SapiensAnatis:
Reverts SapiensAnatis/Dawnshard#444

There have been a bunch of errors in the logs today about players being short of materials which I've never seen before. In other words, the client is out of sync with the backend about what materials it owns.

I'm not quite clear yet on why disabling change tracking might cause this, but it seems the most likely candidate. The only one that seems relevant is LoadService, and that query is after a SaveChangesAsync. Best to revert the changes and investigate the cause before re-applying. Performance benefit was not immediately apparent anyway.


SapiensAnatis:
Doesn't seemed to have fixed some material shortage issues, may be unrelated, need to investigate further

PR #460, opened 2023-10-22
Update TutorialService.cs

FatesTimelines1337:
Should fix people not getting the dojos that come with the imperial onslaught


SapiensAnatis (approved):
Thanks for the pull request -- I would prefer to work out what is blocking tutorial progression past 30102 and there are likely intermediate stages between that and the max status, but it will fix the issues you and other people have had for now.

Issue #397, opened 2023-08-16
Quest drops rework #3


SapiensAnatis:
Following the random drop generation implemented in #301 we should add drop tables for certain quests to make their drops more realistic.

Issue #352, opened 2023-08-01
Add ability to open reward chests on the story screen


SapiensAnatis:
See https://dragalia-api-docs.readthedocs.io/en/latest/dragalia/quest_open_treasure.html

For IDs see QuestTreasureData in masterasset. Rewards documented at https://dragalialost.wiki/w/Main_Campaign#Treasure_Chests

PR #464, opened 2023-10-22
Add Wyrmprints to quest drops

FatesTimelines1337:
- [ ] Add drop rates for prints
- [ ] Add multipliers for prints
- [ ] Test that everything works


SapiensAnatis:
Before you put too much work into this, I should say I actually want to rework the quest drop system quite a bit. This includes #397 so that instead of generating stuff based on material rarity the quest drop list is specified as a list of arbitrary entities. 

This is because if we start tacking on wyrmprints to materials, later we'll have to add eldwater, honey, ashes, blazons, .... and eventually the drop generation is going to look really ugly. It should be completely data driven IMO

FatesTimelines1337:
Yeah, that sounds like a better idea

SapiensAnatis:
@FatesTimelines1337 if you are still interested in contributing, #352 is a bit of a smaller task that would still be appreciated!

PR #467, opened 2023-10-23
Open Quest Treasure Chests

FatesTimelines1337:
To do:
- [x] Add Treasure Data for all other chapters
- [x] Find item ids for Exquisite Honey and Summon Voucher
- [x] Find out to get all the data for the response


FatesTimelines1337:
Theoretically, this should close #352. Testing on a local server still needs to be done through.

FatesTimelines1337:
### Items to send:
- [ ] update_data_list, should have treasure id in quest treasure list
- [ ] entity_result
- [ ] quest_treasure_reward_list, should this be empty for dragon storage reward?
- [ ] duplicate_entity_list, how is this built for this purpose?
- [ ] add_max_dragon_quantity
- [ ] add_max_weapon_quantity, should always be 0
- [ ] add_max_amulet_quantity, should always be 0

SapiensAnatis:
You don't have to bother with dragon storage for now if you don't want -- since dragon storage doesn't do anything on Dawnshard. Equally you can implement it if you want, I believe you will need to update a column in DbPlayerUserData.

It may not have an entity type, since it isn't a tangible 'thing'. The list we have in the enum was datamined so if it's not there you probably won't find it.

To get the update_data_list working, you will probably need to create a new DB table for claimed quest treasure rewards and add this to the UpdateDataService which will cause changes in it to be detected

DragaliaAPI/Features/Quest/QuestController.cs
```diff
@@ -107,6 +126,71 @@ QuestGetQuestClearPartyMultiRequest request
         );
     }
 
+    [HttpPost("open_treasure")]
```
SapiensAnatis, line 129: Please can you add some automated tests of this endpoint to validate your changes?

SapiensAnatis (changes requested):
Good work on this initial implementation, I think it should work. I'd like it to be polished up a bit before merging in though.

DragaliaAPI.Database/Entities/DbQuestTreasureList.cs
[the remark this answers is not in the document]
FatesTimelines1337: It's having trouble with ApiContext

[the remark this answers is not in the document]
SapiensAnatis: try `dotnet ef migrations add <name> -p DragaliaAPI.Database -s DragaliaAPI`

DragaliaAPI.Test/Controllers/QuestControllerTest.cs
```diff
@@ -112,4 +115,116 @@ public async Task ReadStory_ProducesExpectedResponse()
         this.mockStoryService.VerifyAll();
         this.mockUpdateDataService.VerifyAll();
     }
+
+    [Fact]
+    public async Task OpenTreasure_ProducesExpectedResponse()
```
SapiensAnatis, line 120: I'd prefer a test in the integration test project that tests the whole thing rather than just the controller if possible

DragaliaAPI.Database/Entities/DbQuestTreasureList.cs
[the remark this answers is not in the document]
FatesTimelines1337: It's throwing a `no such host is known` error when I use those flags to update the database.

FatesTimelines1337:
I don't understand what's happening, but for some reason Quest Controller stopped working

FatesTimelines1337:
Found the cause, forgot to add the service to the config

FatesTimelines1337:
Not entirely sure if I'm checking everything with the returned response, but the api should be working now.

DragaliaAPI/Services/Game/QuestTreasureService.cs
```diff
@@ -0,0 +1,82 @@
...
+            request.quest_treasure_id
+        ];
+
+        await rewardService.GrantReward(
+            new Entity(
+                questTreasureData.EntityType,
+                questTreasureData.EntityId,
+                questTreasureData.EntityQuantity
+            )
+        );
```
SapiensAnatis, line 34: What does this do when EntityType/EntityId/EntityQuantity are 0, like for `102302`?

SapiensAnatis:
I also just remembered `quest_treasure_list` is in the savefile and needs to be returned from `/load/index`. Could you add it to the save import and `LoadService` please?

SapiensAnatis:
> Not entirely sure if I'm checking everything with the returned response, but the api should be working now.

It's looking a lot better! Thanks for working on this. Just the two details above and then I'm happy to merge this in.

DragaliaAPI/Services/Game/QuestTreasureService.cs
FatesTimelines1337: zero is Entity Type None so rewardService isn't used in that case.

SapiensAnatis (approved):

PR #481, opened 2023-11-04
StoryService : Add adventurer story epithet rewards

LatiWednesday:


DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -134,25 +134,33 @@ int storyId
...
     {
         UnitStory data = MasterAsset.UnitStory[storyId];
-
+        StoryData charaStory = MasterAsset.CharaStories[data.ReleaseTriggerId];
+        List<AtgenBuildEventRewardEntityList> rewardList = new();
         DbPlayerStoryState story = await storyRepository.GetOrCreateStory(data.Type, storyId);
-
         int wyrmiteReward;
+        int emblemRewardId;
+        Entity emblemRewardEntity;
```
SapiensAnatis, line 142: why do these need to be declared out here?

DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -134,25 +134,33 @@ int storyId
...
 
         if (data.Type == StoryTypes.Chara)
+        {
             wyrmiteReward = data.IsFirstEpisode ? CharaStoryWyrmite1 : CharaStoryWyrmite2;
+            if (data.Id == charaStory.storyIds.Last())
+            {
+                emblemRewardId = data.ReleaseTriggerId;
+                emblemRewardEntity = new Entity(EntityTypes.Title, emblemRewardId, 1);
+                await rewardService.GrantReward(emblemRewardEntity);
+                rewardList.Add(new() { entity_type = EntityTypes.Title, entity_id = emblemRewardId, entity_quantity = 1 });
```
SapiensAnatis, line 152: You can use `.ToBuildEventRewardEntityList` on the emblem entity

DragaliaAPI.Test/Services/StoryServiceTest.cs
```diff
@@ -195,6 +195,28 @@ public async Task ReadUnitStory_StoryRead_ReturnsExpectedReward(StoryTypes type,
         this.mockStoryRepository.VerifyAll();
     }
 
+    [Theory]
+    [InlineData(StoryTypes.Chara, 100001085, 10150403)]
```
SapiensAnatis, line 199: Could you add some more cases? Not much point of a `[Theory]` with only one case

DragaliaAPI/Services/Game/StoryService.cs
SapiensAnatis, line 1: You're missing the savefile update that we talked about -- we should grant epithets to everyone who has already met the requirement for them, otherwise these players will never be able to get them

SapiensAnatis (changes requested):
Thank you for your work on this. It looks like this should work. Just a few small comments about the implementation.

SapiensAnatis:
re the lint pipeline, the code is formatted with [CSharpier](https://marketplace.visualstudio.com/items?itemName=csharpier.csharpier-vscode) to ensure a consistent style

DragaliaAPI/Services/Game/StoryService.cs
LatiWednesday: Force of habit, I'm used to declaring all my variables at the beginning of the function, would you rather have me declare them only in the path where I use them ?

SapiensAnatis: On second thought, it's only a style preference -- fine to leave as is

DragaliaAPI/Features/SavefileUpdate/V13Update.cs
```diff
@@ -0,0 +1,42 @@
...
+    public V13Update(
+        IStoryRepository storyRepository,
+        IEmblemRepository emblemRepository,
+        ILogger<V13Update> logger
+    )
+    {
+        this.storyRepository = storyRepository;
+        this.emblemRepository = emblemRepository;
+        this.logger = logger;
+    }
```
LukeFZ, line 23:
```suggestion
public class V13Update(
    IStoryRepository storyRepository,
    IEmblemRepository emblemRepository,
    ILogger<V13Update> logger
) : ISavefileUpdate
```

DragaliaAPI/Features/SavefileUpdate/V13Update.cs
```diff
@@ -0,0 +1,42 @@
+using DragaliaAPI.Database.Repositories;
+using DragaliaAPI.Features.Emblem;
+using DragaliaAPI.Shared.Definitions.Enums;
+using DragaliaAPI.Shared.MasterAsset;
+
```
LukeFZ, line 5:
```suggestion
using Microsoft.EntityFrameworkCore;

```

DragaliaAPI/Features/SavefileUpdate/V13Update.cs
```diff
@@ -0,0 +1,42 @@
...
+        this.storyRepository = storyRepository;
+        this.emblemRepository = emblemRepository;
+        this.logger = logger;
+    }
+
+    public int SavefileVersion => 13;
+
+    public async Task Apply()
+    {
+        int[] readStoryIdList = storyRepository.UnitStories.Where(x => (x.State == StoryState.Read) && x.StoryType == StoryTypes.Chara).Select(x => x.StoryId).ToArray();
```
LukeFZ, line 29:
```suggestion
        int[] readStoryIdList = await storyRepository.UnitStories.Where(x => (x.State == StoryState.Read) && x.StoryType == StoryTypes.Chara).Select(x => x.StoryId).ToArrayAsync();
```

DragaliaAPI/Features/SavefileUpdate/V13Update.cs
```diff
@@ -0,0 +1,42 @@
...
+        int storyCharacterId;
+        int[] characterStoryList;
+        foreach (int readStoryId in readStoryIdList)
+        {
+            storyCharacterId = MasterAsset.UnitStory[readStoryId].ReleaseTriggerId;
+            characterStoryList = MasterAsset.CharaStories[storyCharacterId].storyIds;
+            if (readStoryId == characterStoryList.Last())
```
LukeFZ, line 36:
```suggestion
            if (readStoryId == characterStoryList[-1])
```
Not sure if this is better for perf but it doesnt use the IEnumerable func

LukeFZ (changes requested):
good imo, just a few minor things
(ps. unsure if this is the optimal way to query the stories, maybe getting all last chara stories the n querying based on that is faster? depends on how many unitstories there are that dont match)

DragaliaAPI/Features/SavefileUpdate/V13Update.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                emblemRepository.AddEmblem((Emblems)storyCharacterId);
```
SapiensAnatis, line 38: You will need to check that it isn't already owned as someone could have a story emblem if it's their `emblem_id` in their save, which would cause this to crash on a PK violation. Or you could just use RewardService which handles the check for you

DragaliaAPI/Features/SavefileUpdate/V13Update.cs
SapiensAnatis: Would EF be able to compile a `.Where(x => x.Id % 10 == 5)`? Looks like all the story ids end in 1, 2, 3, 4, or 5.

LukeFZ: doubt it imo, I think an optimization would be to just precalculate all last chara stories then query based on that array/list + state

LatiWednesday: Is it really that worth it to optimize here ? Given that it's only performed once per save file

SapiensAnatis (dismissed):

SapiensAnatis:
Happy with the changes, will merge once you can get the tests/pipelines sorted out

LukeFZ (approved):

SapiensAnatis (approved):
Will merge tomorrow at a more reasonable time for me to do an announcement :) thanks for your work on this

PR #501, opened 2023-11-19
Fix save import time not updating

SapiensAnatis:
Seems to have been broken in #500 going by the latest save import time in the db, but can't find where this was being done before then either 😲


SapiensAnatis:
found it https://github.com/SapiensAnatis/Dawnshard/blob/84f00fffc6f7d64a0bd0d832676ea529bb150e2b/DragaliaAPI/Services/Game/SavefileService.cs#L177

was broken when I removed that AfterMap call

Issue #499, opened 2023-11-15
Mercurial gauntlet doesn't trigger the reward popup


SapiensAnatis:
**Describe the bug**
Mercurial gauntlet doesn't trigger the reward popup on the 15th of every month.

**Expected behavior**
Upon logging in on the 15th of the month after reset, a popup showing MG rewards can be claimed should be visible, if the user has unlocked MG and made progress in it. Clicking this should grant the reward

**Additional context**
This is driven by `monthly_wall_reward_list` from the response given by `login/index`. Set the `quest_group_id` equal to the MG quest group and the reward status to `Available` and the popup will appear on login.

SapiensAnatis:
Closed by #797

Issue #507, opened 2023-11-22
T3 event materials unavailable in Accursed Archives and Dream Big Under the Big Top


SapiensAnatis:
**Describe the bug**
Tier 3 event materials (Colorful Bouquet / Divulgent Lenses) don't drop from challenge battles in the compendium versions of 'Accursed Archives' and 'Dream Big Under the Big Top'

**Expected behavior**
These materials should be available from challenge battles

**Additional context**
This is because in `EventDropService` we subtract the quest's `VariationType` from `VariationTypes.Hard` as a means of limiting T3 drops to challenge battles:

```csharp
// T2 and T3 currency are only used for redeeming shop items, and can appear commonly from Challenge Battle quests.
if (type == DungeonTypes.Wave)
{
    // T3 only drops from Challenge Battle quests
    int t3Quantity = GenerateDropAmount(
        10 * record.wave * ((variation - VariationTypes.Hard) / 2d) * buildDropMultiplier
    );
    yield return new Entity(evt.ViewEntityType3, evt.ViewEntityId3, t3Quantity);
}
```
If a drop with <0 quantity is yielded it will not be awarded.

This works for newer events, but these two events (and possibly others) have all of their challenge battles set to `VariationTypes.Normal` so no quests are eligible for T3 drops.

A proposed fix would be to change the `(variation - VariationTypes.Hard) / 2d` logic to a `GetIsEligibleForT3()` call or similar which can implement a special case of logic for these two events and any others that may be affected. We should ideally remove any dependency on `VariationType` even for quantity due to this discrepancy so that some events aren't harder to grind than others.

SidMalladi:
@SapiensAnatis Hi Jay, can I please be assigned to this issue?

I need a PR merged into an open-source code as part of my assignment for a class, so I'll work on this quickly.

SapiensAnatis:
Looking at the VariationType of wave quests (DungeonType 6) we see there's quite a range between events:

```sql
SELECT t._Text as EventName, q._Id as QuestId, q._VariationType as QuestVariationType FROM EventData e
JOIN QuestData q on e._Id = q._Gid
JOIN TextLabel t on t._Id = e._Name
WHERE e._IsMemoryEvent = 1
AND q._DungeonType = 6
```


EventName | QuestId | QuestVariationType
-- | -- | --
Flames of Reflection | 208160501 | 3
Flames of Reflection | 208160502 | 4
The Miracle of Dragonyule | 208170501 | 1
[...]

PR #509, opened 2023-11-23
Mercurial gauntlet

SidMalladi:
#499 Mercurial gauntlet doesn't trigger the reward popup

### DIFF
--- a/DragaliaAPI/Features/Event/EventDropService.cs
+++ b/DragaliaAPI/Features/Event/EventDropService.cs
@@ -165,6 +165,19 @@ public class EventDropService(IRewardService rewardService, IEventRepository eve
         return dropList;
     }
 
+    private bool IsEligibleForT3Drops(QuestData quest, EventData evt)
+{
+    // Check for specific event names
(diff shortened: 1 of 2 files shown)


SapiensAnatis:
You appear to have just implemented validation on the endpoint to claim a reward. While that is probably something we should do, this is not really what the issue meant, it was referring to sending a notification that a reward is available. 

Additionally, the reward does not have to be claimed on the day of the 15th so we will need a new table to track every player's last claim date to work out if they are eligible. 

This will be a bit more complicated than your fix for #507 so maybe for an initial contribution it would be best to focus on that one?

SidMalladi:
> You appear to have just implemented validation on the endpoint to claim a reward. While that is probably something we should do, this is not really what the issue meant, it was referring to sending a notification that a reward is available.
> 
> Additionally, the reward does not have to be claimed on the day of the 15th so we will need a new table to track every player's last claim date to work out if they are eligible.
> 
> This will be a bit more complicated than your fix for #507 so maybe for an initial contribution it would be best to focus on that one?

Hi @SapiensAnatis , can you check if #507  is good to be merged? I've made the requested changes.

SapiensAnatis:
Closing due to inactivity

PR #514, opened 2023-11-28
Refactor MasterAsset to use FrozenDictionary and preload it on startup

LukeFZ:


DragaliaAPI/Program.cs
```diff
@@ -24,6 +27,15 @@
...
 WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
 
+Stopwatch watch = new();
+Console.WriteLine("Loading MasterAsset data.");
+
+watch.Start();
+RuntimeHelpers.RunClassConstructor(typeof(MasterAsset).TypeHandle);
+watch.Stop();
+
+Console.WriteLine($"Loaded MasterAsset in {watch}.");
```
SapiensAnatis, line 37: Could move this down to after we build the app and use app.logger -- then it will go to seq and you can have structured logging. 

Not that important for a minor performance diagnostic I suppose

DragaliaAPI.Shared/MasterAsset/MasterAsset.cs
```diff
@@ -265,25 +265,25 @@ public static class MasterAsset
     /// <summary>
     /// Dragon StoryId Arrays indexed by DragonId
     /// </summary>
-    public static MasterAssetData<int, StoryData> DragonStories { get; } =
+    public static readonly MasterAssetData<int, StoryData> DragonStories =
```
SapiensAnatis, line 268: Could we make it consistent to use get-only public properties instead of fields? I don't think it matters that much since we won't be using custom get set accessors but just better to have public members as properties imo

DragaliaAPI.Shared/MasterAsset/MasterAssetData.cs
```diff
@@ -80,44 +81,42 @@ public MasterAssetData(string jsonFilename, Func<TItem, TKey> keySelector)
...
-
-        return result;
+        return new FrozenKeyedCollection(items, this.keySelector);
     }
 
-    private class InternalKeyedCollection : KeyedCollection<TKey, TItem>
+    private class FrozenKeyedCollection
     {
-        private readonly Func<TItem, TKey> keySelector;
+        public readonly ImmutableArray<TItem> Items;
```
SapiensAnatis, line 108: Same comment, would prefer this as a get-only property

DragaliaAPI/Program.cs
LukeFZ: addressed

PR #524, opened 2023-12-06
Fix Wyrmite payment calculation for finishing fort builds

LukeFZ:
Until now this was based on the buildStartTime, but it should actually be calculated based on the current time.


SapiensAnatis (dismissed):
Thanks 😄

SapiensAnatis (approved):

Issue #613, opened 2024-01-19
Kaleidoscape cannot drop Hitter portrait prints


sockperson:
https://github.com/SapiensAnatis/Dawnshard/blob/405a8180fcd49c6c1ea32c1e2349b192c34db8cc/DragaliaAPI/Features/DmodeDungeon/TalismanHelper.cs#L116C9-L116C58

Shiny/Klaus observed that the Kaleidoscape wasn't dropping portrait prints with Hitter abilities.

`int index = rdm.Next(startIndex, pool.Count) - 1;` (TalismanHelper.cs:116)

Could this be because the randomly chosen index for the ability pool can only be a maximum of 18 (max(rdm.Next(2, 20) - 1)) = 18), when the Hitter abilities are index 19 in `Ability2Pool`?

SapiensAnatis:
I think that is probably correct, it looks like this code doesn’t account for the fact that rdm.Next will never return a value equal to pool.Count and so the -1 is not necessary. 

Beyond that, a 1/19 chance for a hitter ability seems too low. I would welcome a pull request to adjust this drop logic.

PR #614, opened 2024-01-19
Fix GenerateAbility ability index

sockperson:
Fix random index to be able to choose max index (pool.Count - 1) rather than (pool.Count - 2)

Fixes https://github.com/SapiensAnatis/Dawnshard/issues/613


SapiensAnatis (approved):
Thanks

PR #654, opened 2024-02-18
Fix Random.Next collection extensions being unable to return their last element

LukeFZ:


SapiensAnatis:
@LukeFZ Changes look good, but I was just reviewing some other uses of `.Next` and found this:

https://github.com/SapiensAnatis/Dawnshard/blob/c17b7f482cc5f2643ed86f9d84977ebbbde15c55/DragaliaAPI/DragaliaAPI/Features/DmodeDungeon/DmodeDungeonService.cs#L178

I think this suffers from the same problem with the exclusive upper bound -- is this not basically equivalent to

```cs
int talismanAmount = floorNum switch
{
    < 50 => 1,
    < 60 => 3,
    _ => 6
};
```

If so could you adjust that as part of this please?

LukeFZ:
Fixed all remaining Random.Next calls, including that talisman one

SapiensAnatis (approved):
Thanks!

Issue #645, opened 2024-02-14
KeyNotFound errors in photon plugin


SapiensAnatis:
Following error seen in logs:

```
2024-02-14 07:36:39,698 [67] ERROR Photon.Hive.HiveGame.HiveHostGame.Plugin - UnhandledException:'Name:GluonPlugin, Version:1.0 AppId:B8DB-8B89-AA5D-D998-36C6-FBF0, AppVersion:'. Exception: System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.ThrowHelper.ThrowKeyNotFoundException()
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at DragaliaAPI.Photon.Plugin.Plugins.GameLogic.GameLogicPlugin.RaisePartyEvent()
   at DragaliaAPI.Photon.Plugin.Plugins.GameLogic.GameLogicPlugin.OnSetGoToIngameState(ICallInfo info)
   at DragaliaAPI.Photon.Plugin.Plugins.GameLogic.GameLogicPlugin.OnLeave(ILeaveGameCallInfo info)
   at DragaliaAPI.Photon.Plugin.Plugins.Gluon.GluonPlugin.OnLeave(ILeaveGameCallInfo info)
   at Photon.Hive.Plugin.PluginBase.Photon.Hive.Plugin.IGamePlugin.OnLeave(ILeaveGameCallInfo info)
   at Photon.Hive.Plugin.PluginWrapper.OnLeave(ILeaveGameCallInfo info)
```

Most likely related to hero param processing after a player has left, GameActors may not be updated and iterating through each of them tries to use keys that have been removed from the actor state.

Try to wrap all indexes of actorstate in this method and afterwards with TryGetValue

PR #666, opened 2024-02-23
Fix a few occurrences of KeyNotFoundException in the plugin

SapiensAnatis:
Most of these are to do with the logic to bring the minimum GoToIngameState forward when an actor leaves. Typically the cause is iterating through `PluginHost.GameActors` and then indexing each element's `ActorNr` into a state dictionary... we remove the actor from the state dictionary on leave, but the actor is not always removed from the plugin host list at the same time.

Additionally:
- Refactor all `GoToIngameState` logic into `GoToIngameStateManager` to make `GameLogicPlugin` more manageable
- Change the raising of the character data event into `GoToIngameState = 4`, based on the enum in the il2cpp.dll specifying this value:

  ```cs
  public enum GoToIngameState
  {
	  None,
	  TryGoToIngame,
	  PreCheckParty,
	  CheckParty,
	  PostCheckParty
  }
  ```

### DIFF
--- a/PhotonPlugin/DragaliaAPI.Photon.Plugin/Shared/Helpers/PluginHostExtensions.cs
+++ b/PhotonPlugin/DragaliaAPI.Photon.Plugin/Shared/Helpers/PluginHostExtensions.cs
@@ -1,0 +1,97 @@
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using DragaliaAPI.Photon.Plugin.Plugins.GameLogic.Events;
+using DragaliaAPI.Photon.Plugin.Shared.Constants;
+using MessagePack;
+using Photon.Hive.Plugin;
+
+namespace DragaliaAPI.Photon.Plugin.Shared.Helpers
+{
+    public static class PluginHostExtensions
+    {
+        private const int EventDataKey = 245;
+        private const int EventActorNrKey = 254;
+
+        private static readonly MessagePackSerializerOptions MessagePackOptions =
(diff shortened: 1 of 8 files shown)


SapiensAnatis:
Closes #645 - will reopen if issues still present

Shared/DragaliaAPI.Photon.Shared/Models/HeroParamResponse.cs
```diff
@@ -6,6 +6,10 @@ public class HeroParamData
 
         public long ViewerId { get; set; }
 
+        public int UsedMemberCount { get; set; }
+
```
SapiensAnatis, line 10: Bad

PhotonPlugin/DragaliaAPI.Photon.Plugin/Shared/Helpers/PluginHostExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+            return pluginHost.GameProperties.GetInt(GamePropertyKeys.QuestId);
+        }
+
+        /// <summary>
+        /// Logs an error if a HTTP response was not successful.
+        /// </summary>
+        /// <param name="pluginHost">Instance of <see cref="IPluginHost"/>.</param>
+        /// <param name="httpResponse">The HTTP response.</param>
+        /// <param name="userState">The user state.</param>
+        public static void LogIfFailedCallback(
```
SapiensAnatis, line 81: Use elsewhere

PR #683, opened 2024-02-26
Increase CustomWebApplicationFactory lifetime

SapiensAnatis:
Use it via an ICollectionFixture instead of IClassFixture, so that one instance lives for the entire test run. Should speed up the integration tests a bit.


SapiensAnatis:
Didn't actually speed up the tests by much, but does make the test fixture code a tad cleaner.

PR #777, opened 2024-04-29
Campaign skip feature

jaredstrong89:
I have implemented the campaign skip (story skip) functionality. When the "Skip Campaign" button is clicked on the client application (only visible when player has not cleared chapter 10), player data will be adjusted as described in the Dragalia Lost Wiki.

- Player level: 60
- Stamina: 999
- Getherwings: 99
- All quest first clear and endeavors completed on Easy mode from chapters 1-10, granting Wyrmite rewards
- All quest stories marked as "read" from chapters 1-10
- Story units and dragon rewards added to collection
- Tutorials skipped through chapter 10
- Story progression Halidom facilities added to storage with their appropriate levels

I also included a fairly detailed unit test, which will check player level, Halidom buildings, etc.

[Api Doc](https://dragalia-api-docs.readthedocs.io/en/latest/dragalia/story_skip_skip.html)
[Dragalia Lost Wiki](https://dragalialost.wiki/w/Beginner%27s_Guide#The_Story_Skip_Option)

This was my first dive into a C# web app, so hopefully my code fits with your coding conventions.


DragaliaAPI/DragaliaAPI/Features/StorySkip/IStorySkipService.cs
```diff
@@ -0,0 +1,10 @@
+namespace DragaliaAPI.Features.Quest;
+
+public interface IStorySkipService
```
SapiensAnatis, line 3: I would get rid of this interface - I know most services in the codebase have an interface, but I've stopped writing them recently as they don't add much value if you only have one implementation.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,191 @@
...
+        );
+
+        EntityResult entityResult = StoryService.GetEntityResult(rewardList);
+        IEnumerable<AtgenQuestStoryRewardList> questRewardList = rewardList.Select(
+            StoryService.ToQuestStoryReward
+        );
+
+        UpdateDataList updateDataList = await updateDataService.SaveChangesAsync(
+            cancellationToken
+        );
```
SapiensAnatis, line 188: Why do you need this `SaveChanges`? Calling it in a loop will be very expensive

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,191 @@
...
+            );
+        }
+
+        IEnumerable<BuildList> buildListEnum = await fortService.GetBuildList();
+        foreach (BuildList buildList in buildListEnum)
+        {
+            if (fortConfigs.TryGetValue(buildList.PlantId, out FortConfig fortConfig))
+            {
+                DbFortBuild build = await fortRepository.GetBuilding((long)buildList.BuildId);
+                if (build.Level < fortConfig.Level)
```
SapiensAnatis, line 98: Would it be better to use `FortRepository.Builds` here, to skip the extra lookup, and to do the filtering as part of the DB query? This might work:
```suggestion
        List<DbFortBuild> buildListEnum = await fortRepository.Builds.Where(x => fortConfigs.Keys.Contains(x.PlantId)).ToListAsync();
        foreach (BuildList buildList in buildListEnum)
        {
            fortConfig = fortConfigs[build.PlantId];
            if (build.Level < fortConfig.Level)
```

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,191 @@
...
+            data.Exp = MaxExp;
+        }
+
+        if (data.Level < MaxLevel)
+        {
+            data.Level = MaxLevel;
+        }
+    }
+
+    public async Task OpenTreasure(int questTreasureId, CancellationToken cancellationToken)
```
SapiensAnatis, line 129: Do we have to open treasure as part of the story skip? The wiki doesn't mention it anywhere

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,191 @@
...
+                {
+                    build.Level = fortConfig.Level;
+                    build.BuildStartDate = DateTimeOffset.UnixEpoch;
+                    build.BuildEndDate = DateTimeOffset.UnixEpoch;
+                }
+            }
+        }
+    }
+
+    public async Task IncreasePlayerLevel()
```
SapiensAnatis, line 108: Does a bit more than increase level, maybe best to rename?
```suggestion
    public async Task UpdateUserData()
```

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    public async Task IncreaseFortLevels()
+    {
+        Dictionary<FortPlants, FortConfig> fortConfigs = FortConfigurations.FortConfigs;
+
+        foreach (KeyValuePair<FortPlants, FortConfig> keyValuePair in fortConfigs)
+        {
+            FortPlants fortPlant = keyValuePair.Key;
+            FortConfig fortConfig = keyValuePair.Value;
```
SapiensAnatis, line 83: Nitpick: you can destructure here to make this cleaner

```suggestion
        foreach ((FortPlants fortPlant, FortConfig fortConfig) in fortConfigs)
        {
```

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,191 @@
...
+        bool[] newState = { true, true, true };
+
+        QuestRewardData rewardData = MasterAsset.QuestRewardData[questId];
+        for (int i = 0; i < 3; i++)
+        {
+            (QuestCompleteType type, int value) = rewardData.Missions[i];
+
+            if (!currentState[i])
+            {
+                newState[i] = true;
```
SapiensAnatis, line 44: I don't think this does anything

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipController.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            IEnumerable<QuestData> questDatas = MasterAsset.QuestData.Enumerable.Where(x =>
+                x.Gid < 10011 && x.Id > 100000000 && x.Id.ToString().Substring(6, 1) == "1"
+            );
+            IEnumerable<QuestTreasureData> questTreasureDatas = MasterAsset.QuestTreasureData.Enumerable.Where(x =>
+                x.Id is < 110000 && x.Id.ToString().Substring(3, 1) == "1"
+            );
+            IEnumerable<QuestStory> questStories = MasterAsset.QuestStory.Enumerable.Where(
+                x => x.GroupId is < 10011
+            );
```
SapiensAnatis, line 51: It may be worth caching these in a `FrozenSet` as a private field in `StorySkipService` so that they don't have to  be recalculated for each request -- they won't change after all

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipController.cs
```diff
@@ -0,0 +1,88 @@
...
+    public async Task<DragaliaResult> Read(
+        CancellationToken cancellationToken
+    )
+    {
+        try
+        {
+            this.logger.LogDebug("Beginning story skip for player.");
+
+            IEnumerable<QuestData> questDatas = MasterAsset.QuestData.Enumerable.Where(x =>
+                x.Gid < 10011 && x.Id > 100000000 && x.Id.ToString().Substring(6, 1) == "1"
```
SapiensAnatis, line 44: Why do you need the filter on `Id > 100000000` and that the 6th digit is `1`?

SapiensAnatis (changes requested):
Thank you for this! A very good first attempt for someone new to the codebase and C# as well. I haven't yet pulled this down to test it locally but just some initial feedback.

I think that we may want some specialist infrastructure here to avoid making hundreds of DB calls - yes story skip is only once per account and boots you to the title screen, but it will stress the DB if people do it quite frequently. I'm mainly concerned here about some of the 'get single entity' calls in the loops -- for instance `ReadStory` I think  will call `FindAsync` 104 times and `GetQuestDataAsync` 114 times.

I'll have a think about what that specialist infrastructure could look like - usually we just preload stuff manually and skip the otherwise async/query checks via a dict lookup, but maybe we could use a merge statement or similar. Hope this is at least something to start on.

SapiensAnatis:
P.S. csharpier is an auto-formatter, to get it to pass run
```bash
dotnet tool restore
dotnet csharpier .
```
from the repository root. I'm also happy to run it before merging if you'd prefer that.

jaredstrong89:
> P.S. csharpier is an auto-formatter, to get it to pass run
> 
> ```shell
> dotnet tool restore
> dotnet csharpier .
> ```
> 
> from the repository root. I'm also happy to run it before merging if you'd prefer that.

These tools are new to me, so this is good to know. Thank you!

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipController.cs
jaredstrong89: While comparing the different quest Ids, I found that index 6 was a flag for the easy mode missions. I tested this out, and it worked nicely.
1 = easy, 2 = hard, 3 = very hard
The chapter number and quest number are encoded in there as well.

Because I was doing the substring, I was hitting a string length error on that Id. Instead of investigating further, I filtered for Id > 100000000.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipController.cs
jaredstrong89: I will look into this.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: Yes, I think you're right. I copy/pasted this from the mission completion service. I made just the essential changes, but I can clean it up.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: This is good to know! I'll try it out.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: Noted.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: This was one of those nuances that I was unsure about. I can remove the treasure opening.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: That does sound simpler... Noted.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: Maybe you can help me understand database updates better. Should I only call the update service at the end after all data changes were made?

DragaliaAPI/DragaliaAPI/Features/StorySkip/IStorySkipService.cs
jaredstrong89: Ah ok, that sounds great. Yeah most of the larger files had interfaces, so I followed suit.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipController.cs
```diff
@@ -0,0 +1,88 @@
...
+                new StorySkipSkipResponse() { ResultState = 1 }
+            );
+        }
+        catch (Exception e)
+        {
+            this.logger.LogError("Error occurred while skipping story: {Message}", e.Message);
+            return this.Ok(
+                new StorySkipSkipResponse() { ResultState = 0 }
+            );
+        }
```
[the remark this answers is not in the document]
jaredstrong89: Ok, that sounds good.

jaredstrong89:
> Thank you for this! A very good first attempt for someone new to the codebase and C# as well. I haven't yet pulled this down to test it locally but just some initial feedback.
> 
> I think that we may want some specialist infrastructure here to avoid making hundreds of DB calls - yes story skip is only once per account and boots you to the title screen, but it will stress the DB if people do it quite frequently. I'm mainly concerned here about some of the 'get single entity' calls in the loops -- for instance `ReadStory` I think will call `FindAsync` 104 times and `GetQuestDataAsync` will call `FindAsync` 114 times.
> 
> I'll have a think about what that specialist infrastructure could look like - usually we just preload stuff manually and skip the otherwise async/query checks via a dict lookup, but maybe we could use a merge statement or similar. Hope this is at least something to start on.

Thank you for all the feedback! You provided lots of ideas for me to think about. I have spent some of my free time in the past 2 weeks learning how this repo works, figuring out how to develop .NET on a Mac, with little C# experience and no Postgres knowledge. It's been a fun challenge.

When I got started writing the code, my strategy was to tie in to the other services to be thorough with granting rewards (like story characters and dragons), trying to not miss a step in each of the processes. But I see your point. We definitely want to be less chatty with the database.

Now that I can wrap my head around this thing much better, I have a better understanding of the database and how it relates to the client application. Going forward, I could pull the necessary data at one time, eliminating multiple lookups per table. I could remove references to those other services that handle updates, performing table updates/inserts directly. I will probably have to set up more lists, such as story character unlocks.

jaredstrong89:
Lots of changes with this last commit. There should be way less database lookups. And I ran CSharpier this time!

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    public int ProcessQuestCompletions(long viewerId)
+    {
+        int wyrmite = 0;
+        FrozenSet<QuestData> questDatas = MasterAsset
+            .QuestData.Enumerable.Where(x =>
+                x.Gid < 10011 && x.Id > 100000000 && x.Id.ToString().Substring(6, 1) == "1"
+            )
+            .ToFrozenSet();
```
SapiensAnatis, line 80: This will still get computed on every execution of this method. You need to hoist it as a `static readonly` variable in the class itself so that it is shared between instances:

```cs
public class StorySkipService
{
    private static readonly FrozenSet<QuestData> // ...
```

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
```diff
@@ -0,0 +1,304 @@
...
+    IStoryRepository storyRepository,
+    IUserDataRepository userDataRepository
+)
+{
+    public void IncreaseFortLevels(long viewerId)
+    {
+        Dictionary<FortPlants, FortConfig> fortConfigs = StorySkipRewards.FortConfigs;
+        IQueryable<DbFortBuild> userForts = fortRepository.Builds.Where(x =>
+            x.ViewerId == viewerId
+        );
```
SapiensAnatis, line 29: Not sure if you realize, but every time you do a `Any()`, `Count()` or iterate through with `foreach` on an `IQueryable` you are executing a database query. I would suggest realizing the results of the query before you do your processing:

```suggestion
        List<DbFortBuild> userForts = await fortRepository.Builds.Where(x =>
            x.ViewerId == viewerId
        ).ToListAsync();
```

SapiensAnatis:
> Lots of changes with this last commit. There should be way less database lookups. And I ran CSharpier this time!

Glad to hear it. I think you are getting there but perhaps the `IQueryable` stuff has tripped you up a little. It's looking a lot better though.

DragaliaAPI/DragaliaAPI/Features/StorySkip/StorySkipService.cs
jaredstrong89: I did not realize this. I changed these objects to Lists.

SapiensAnatis (approved):
I've done some testing of this locally and it appears to work pretty well! Happy with the code as well.

I'll merge it in tomorrow after work just so I'm on hand in case any issues arise. Thanks once again for working on this.

Issue #787, opened 2024-05-07
Implement chapter 10 non-skip reward


SapiensAnatis:
**Context**
In the original game, if you reached chapter 10 without using the campaign skip option, you were granted the following rewards:

- 69990 player EXP (level 1 -> 60)
- 350 Hustle Hammers
- Enough materials to upgrade the facilities upgraded by campaign skip

**The feature request**
This reward should be implemented on Dawnshard. A savefile update should be implemented to send these rewards to the gift box of players who have already passed the trigger for the reward. It is fine if this ends up granting it to players who have used campaign skip in the window since #525 was implemented. We should ensure that it is not rewarded to players who campaign skip after it is implemented, though.

jaredstrong89:
I'm on it 🙂

SapiensAnatis:
Closed by #788

PR #788, opened 2024-05-08
Feature: chapter 10 completion rewards

jaredstrong89:
This PR addresses [issue #787](https://github.com/SapiensAnatis/Dawnshard/issues/787).

Chapter 10 completion rewards are granted upon reading the last quest story within chapter 10. Hustle hammers and upgrade materials will be sent to the gift box, and 69,990 exp will be granted.

Skipping the campaign will not trigger the chapter 10 rewards.

Re-reading the chapter 10 story will not send completion rewards.


DragaliaAPI/DragaliaAPI.Integration.Test/Dragalia/QuestReadStoryTest.cs
```diff
@@ -82,6 +82,38 @@ public async Task ReadStory_TheLonePaladyn_SetsTutorialStatus()
         response.UpdateDataList.UserData.TutorialStatus.Should().Be(10600);
     }
 
+    [Fact]
+    public async Task ReadStory_Chapter10Completion_GrantsRewards()
+    {
+        await this
+            .ApiContext.PlayerUserData.Where(x => x.ViewerId == this.ViewerId)
+            .ExecuteUpdateAsync(u => u.SetProperty(e => e.Level, 1).SetProperty(e => e.Exp, 0));
```
SapiensAnatis, line 90: I don't think this is realistic, as nobody would be level 1 while completing chapter 10. I would suggest setting level and XP to a pre-known arbitrary value and asserting that your exp afterwards is what you would expect after gaining 69990.

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -343,4 +358,49 @@ AtgenBuildEventRewardEntityList reward
 
         return questReward;
     }
+
+    private static List<Present> GetChapter10PresentList()
+    {
+        Dictionary<Materials, int> materialCounts =
```
SapiensAnatis, line 364: Could hoist this as a FrozenDictionary

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -343,4 +358,49 @@ AtgenBuildEventRewardEntityList reward
...
+        {
+            presents.Add(
+                new Present(
+                    PresentMessage.Chapter10Clear,
+                    EntityTypes.Material,
+                    (int)material,
+                    count
+                )
+            );
+        }
```
SapiensAnatis, line 402: I would recommend taking a more functional approach and using LINQ, it's a bit easier to read and the modern standard for C#.

Instead of accumulating into a list you can do

```cs
List<Present> presents =
    new() { new Present(PresentMessage.Chapter10Clear, EntityTypes.HustleHammer, 0, 350) };

IEnumerable<Present> addedPresents  = materialCounts.Select((key, value) => 
       // Select: transform each element of the sequence
      new Present(
            PresentMessage.Chapter10Clear,
            EntityTypes.Material,
            (int)material,
            count
      )
); 

// The above IEnumerable is lazily evaluated, and realized when added to the list
presents.AddRange(addedPresents);
```

And this can be tidied up even further with C# 12 collection expressions:

```cs
List<Present> presents =
[
    new Present(PresentMessage.Chapter10Clear, EntityTypes.HustleHammer, 0, 350),
    .. materialCounts.Select(
        (key, value) =>
            new Present(PresentMessage.Chapter10Clear, EntityTypes.Material, (int)material, count)
    )
];
```

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -256,6 +261,16 @@ private async Task<List<AtgenBuildEventRewardEntityList>> ReadQuestStory(int sto
             );
         }
 
+        if (storyId == 1001009)
```
SapiensAnatis, line 264: I wonder if it's possible to do all this by extending the generic reward check above? (L218)

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -256,6 +261,16 @@ private async Task<List<AtgenBuildEventRewardEntityList>> ReadQuestStory(int sto
...
+        {
+            logger.LogDebug("Granting chapter 10 completion rewards.");
+
+            IEnumerable<Present> presents = GetChapter10PresentList();
+            presentService.AddPresent(presents);
+
+            await userService.AddExperience(69990);
```
SapiensAnatis, line 271: We are not adding anything to `rewardList` here - does the client correctly show the rewards you have gained?

SapiensAnatis:
Unless I missed it, there isn't a savefile update here to grant the reward to players who have already reached Chapter 10 (and thus can't receive this reward). It was mentioned in #787. Check `Features/SavefileUpdate` for an example -- they're applied in `LoadController` on login.

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
jaredstrong89: I initially added the rewards to QuestStoryRewardInfo.json, but the rewards were presented in the quest reward screen rather than the gift box.

I could remove the Dictionary if I do that if I go this route.
```c#
if (reward.Type == EntityTypes.FortPlant)
{
      await fortRepository.AddToStorage((FortPlants)reward.Id, reward.Quantity, true);
}
else if (storyId == 1001009)
{
      presentService.AddPresent(
            new Present(
                 PresentMessage.Chapter10Clear,
                 (EntityTypes)reward.Type,
                 reward.Id,
                 reward.Quantity
             )
        );
}
else
{
...
``` 
The problem here is that the story reward message mentions that the presents will go to the "Limited" box, but they actually go to the "No Limit" box. I haven't found a way to move these to that box correctly. 🤔

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
jaredstrong89: You're right that they don't show in the reward list. I'll move the code up to the reward section to make sure that happens.

jaredstrong89:
> Unless I missed it, there isn't a savefile update here to grant the reward to players who have already reached Chapter 10 (and thus can't receive this reward). It was mentioned in #787. Check `Features/SavefileUpdate` for an example -- they're applied in `LoadController` on login.

I see what you mean. Sorry, I definitely read that before, but failed to grasp it. I will look into it.

DragaliaAPI/DragaliaAPI/Features/SavefileUpdate/V21Update.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace DragaliaAPI.Features.SavefileUpdate;
+
+/// <summary>
+/// Update to grant chapter 10 completion rewards to players that previously cleared chapter 10 and did not receive rewards.
+/// </summary>
+[UsedImplicitly]
+public class V21Update(
+    ApiContext apiContext,
+    IPresentService presentService,
+    ILogger<V20Update> logger
```
SapiensAnatis, line 21:
```suggestion
    ILogger<V21Update> logger
```

DragaliaAPI/DragaliaAPI/Features/SavefileUpdate/V21Update.cs
```diff
@@ -0,0 +1,73 @@
...
+            logger.LogInformation(
+                "Detected that chapter 10 was completed. Granting completion rewards."
+            );
+
+            if (
+                MasterAsset.QuestStoryRewardInfo.TryGetValue(
+                    StoryId,
+                    out QuestStoryRewardInfo? rewardInfo
+                )
+            )
```
SapiensAnatis, line 54: The result of this lookup should always be the same -- perhaps you could store the `QuestStoryRewardInfo` as a `static readonly` variable?

DragaliaAPI/DragaliaAPI/Services/Game/StoryService.cs
```diff
@@ -217,6 +222,17 @@ private async Task<List<AtgenBuildEventRewardEntityList>> ReadQuestStory(int sto
                 {
                     await fortRepository.AddToStorage((FortPlants)reward.Id, reward.Quantity, true);
                 }
+                else if (storyId == 1001009)
```
SapiensAnatis, line 225: Could you extract this to a named constant or add a comment explaining what it means?

DragaliaAPI/DragaliaAPI/Features/SavefileUpdate/V21Update.cs
SapiensAnatis, line 1: The save update needs to add the XP as well

jaredstrong89: Good catch!

DragaliaAPI/DragaliaAPI.Integration.Test/Other/SavefileImportTest.cs
jaredstrong89, line 1: @SapiensAnatis, I had to add these 3 lines in SaveFileImportTest to pass the unit test, since the v22 adds a large amount of player exp.

Would you handle this a different way?

SapiensAnatis: I think it is okay, we exclude a lot of things there that our server handles in such a way that we can't expect it to be 1:1.

It's probably worth adding a separate set of asserts for those 3 properties if you know what they will be though

DragaliaAPI/DragaliaAPI.Shared/Resources/Story/QuestStoryRewardInfo.json
```diff
@@ -1,617 +1,727 @@
...
+        "_Type": 9,
+        "_Id": 100604,
+        "_Quantity": 1
+      }
+    ]
+  },
+  {
+    "_Id": 1001009,
+    "_Rewards": [
+      {
```
SapiensAnatis, line 170: Did you add these in manually? Where did you get the info from?

DragaliaAPI/DragaliaAPI/Features/SavefileUpdate/V22Update.cs
```diff
@@ -0,0 +1,72 @@
...
+    IPresentService presentService,
+    IUserService userService,
+    ILogger<V22Update> logger
+) : ISavefileUpdate
+{
+    private const int Chapter10LastStoryId = 1001009;
+
+    private static readonly List<QuestStoryReward> Rewards = MasterAsset
+        .QuestStoryRewardInfo.Enumerable.Where(x => x.Id == Chapter10LastStoryId)
+        .First()
```
SapiensAnatis, line 28: Since the ID is the key I would use the O(1) lookup
```suggestion
        .QuestStoryRewardInfo[Chapter10LastStoryId]
```

DragaliaAPI/DragaliaAPI.Shared/Resources/Story/QuestStoryRewardInfo.json
jaredstrong89: Yes, I added the rewards manually. I looked at FortPlantDetail.json to determine the material IDs and count for each level up, dropping the info in a spreadsheet to get the totals.

Is it ok to add to that file?

DragaliaAPI/DragaliaAPI/Features/SavefileUpdate/V22Update.cs
jaredstrong89: Ah this is much nicer!

DragaliaAPI/DragaliaAPI.Shared/Resources/Story/QuestStoryRewardInfo.json
SapiensAnatis: Sure, I think as it has an *Info suffix it was manually generated anyway.

SapiensAnatis (approved):

PR #848, opened 2024-05-30
Fix limit breaks only deducting the materials for the target count

### LINKED ISSUE #802: Incorrect materials in mana circle unlocks
Reported by Sinkarth -

auto-unlock nodes allowed progressing to MC 40 with the following material display:

but then MC 41 was available with this material display:

Note that the game says 32 gold scales will be taken from 513 owned, and 83 silver scales will be taken from 414 owned. Yet in the second screenshot the player ends up with 491 gold scales (-22) and 356 silver scales (-58). In other words fewer scales are deducted than expected.

This indicates in the 1-40 mana nodes upgrades of gala alex that there is a discrepancy between client and server material costs

LukeFZ:
Fixes #802


SapiensAnatis (approved):
Works like a charm on the 40 node Gala Alex upgrade mentioned in the issue. Thank you!

PR #862, opened 2024-06-06
Actually remove dungeon sessions

SapiensAnatis:
~~Speculative fix for random auto-repeat glitches throwing up stale data, and something we should be doing anyway.~~

Today, when you finish a dungeon, the session remains in the cache for the remainder of its expiry time, because issues were previously encountered with the client retrying record requests if they take too long, and then encountering errors because the session had been deleted on the first request.

Now that the codebase makes better use of CancellationToken, this is much less likely if we put the eviction of the session right at the end of the response method. It is still possible if the request is cancelled during serialization, however. If we encounter issues we can make the RemoveDungeon method instead mark the session to be removed in 1 minute's time or similar.

Also includes a refactoring of the DungeonService to implement a unit-of-work pattern, where the current session can be cached and modified in-memory without round-trips to the cache. This is primarily aimed at the DungeonStart logic which can require several updates after creating the session, arising from how the code is structured.

Finally, makes sessions keyed by viewer ID, so that it isn't possible to get another player's session. It probably wasn't really possible before, given the astronomical likelihood of guessing/colliding GUIDs, but it makes sense to segregate this data by tenant all the same. **This key change is breaking, and the server should be taken down during the deployment.**

### DIFF
--- a/DragaliaAPI/DragaliaAPI.Integration.Test/Features/Dungeon/DungeonRecordTest.cs
+++ b/DragaliaAPI/DragaliaAPI.Integration.Test/Features/Dungeon/DungeonRecordTest.cs
@@ -103,7 +103,7 @@ public class DungeonRecordTest : TestFixture
                 }
             };
 
-        string key = await Services.GetRequiredService<IDungeonService>().StartDungeon(mockSession);
+        string key = await this.StartDungeon(mockSession);
 
         DungeonRecordRecordResponse response = (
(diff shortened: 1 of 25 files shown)


SapiensAnatis:
Apparently #861 fixed a lot of the reported issues with auto-repeat, to my surprise. This change still makes sense though as it will clean up the cache and remove an AutoMapper dependency

PR #1017, opened 2024-08-13
FIx result code being logged as null if returning an object directly

SapiensAnatis:
Modern controllers tend to return a `TData` directly, by having a signature of `DragaliaResult<TData>` which returns an `IActionResult` by implementing `IConvertToActionResult`. This bypasses the mechanism implemented in #1016 inside `DragaliaControllerBase` which sets the `ResultCode` based on the use of `Ok()` or `Code()`.

To have the result code set in these cases, use an action filter that runs after all actions and attempts to determine if the result is a `DragaliaResponse` and, if it is, extract the result code. I've had to implement an abstract class for this as it wouldn't be possible to pass a value for the generic argument.

This mechanism replaces any need to set the result code manually in `DragaliaControllerBase`.

Also changed the constructors of `DragaliaResponse` to make them a tad more consistent.


SapiensAnatis:
Consider: could this run as a global action filter (so that it executes last) and replace the middleware?

PR #1107, opened 2024-10-08
Add draconic essence drops

SapiensAnatis:
Add drops of draconic essences to hard / very hard main story quests.

The visual indicator of the number of essence available is driven by `quest_list[].daily_play_count`, which we already handle correctly in `QuestService.ProcessQuestCompletion`. So the only change that needs to be made is checking whether the quest drops essence, and whether our play count is low enough, and then granting it.


SapiensAnatis:
Note: I wondered if the daily reset mechanism inside `ProcessQuestCompletion` needed to be done inside an `IDailyResetAction` instead, so that players could see the correct number of essences before it gets reset by clearing a quest. However, it looks like the game handles this by using the `LastDailyResetTime` :)

PR #1120, opened 2024-10-16
Re-use challengeString cookie if it is already set

SapiensAnatis:
The cause of the login issues appears to be Chrome specifically re-generating the challenge string when it visits /oauth, causing it to not match the hash we sent to BaaS.


SapiensAnatis:
Suddenly stopped being able to repro this even without the fix, but it won't do any harm to handle this case

Issue #668, opened 2024-02-23
Save import fails on first login


SapiensAnatis:
**Describe the bug**
When you log in for the first time ever and have a save to import, the save import fails with

```
Microsoft.EntityFrameworkCore.DbUpdateConcurrencyException: The database operation was expected to affect 1 row(s), but actually affected 0 row(s); data may have been modified or deleted since entities were loaded. See https://go.microsoft.com/fwlink/?LinkId=527962 for information on understanding and handling optimistic concurrency exceptions.
   at Npgsql.EntityFrameworkCore.PostgreSQL.Update.Internal.NpgsqlModificationCommandBatch.ThrowAggregateUpdateConcurrencyExceptionAsync(RelationalDataReader reader, Int32 commandIndex, Int32 expectedRowsAffected, Int32 rowsAffected, CancellationToken cancellationToken)
   at Npgsql.EntityFrameworkCore.PostgreSQL.Update.Internal.NpgsqlModificationCommandBatch.Consume(RelationalDataReader reader, Boolean async, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.ExecuteAsync(IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable`1 commandBatches, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable`1 commandBatches, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.ExecuteAsync(IEnumerable`1 commandBatches, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChangesAsync(IList`1 entriesToSave, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChangesAsync(StateManager stateManager, Boolean acceptAllChangesOnSuccess, CancellationToken cancellationToken)
[...]

PR #1135, opened 2024-10-30
Refactor game authentication code to use ASP.NET JWT Authentication

SapiensAnatis:
Previously, the authentication code manually read the ID token and BaaS JWKs and determined its validity using `JsonWebTokenHandler`. It then threw an exception if the token was invalid, and relied on `ExceptionHandlerMiddleware` to return the appropriate status code for the game.

When I implemented JWT authentication for the website, I used the in-box ASP.NET `AddJwtBearer` authentication and it greatly simplified matters. I have been meaning to do the same for the game authentication code for a while.

I got the final push I needed after #1129 which changed it so that every expired token error was logged as an unhandled exception. This clutters the logs and introduces misleading metrics in Grafana:

image

As part of simplifying the authentication code, I have also removed the legacy authentication flow, which the server used in its first weeks being live before BaaS was established. This means we now have a hard dependency on BaaS, but it is now open source and available for anyone to host. This old authentication flow had not been tested in literal years, so there's a good chance it wasn't functional anyway.


DragaliaAPI/DragaliaAPI/Services/Game/SavefileService.cs
```diff
@@ -436,6 +436,13 @@ private async Task Delete()
     {
         long viewerId = this.playerIdentityService.ViewerId;
 
+        // The usage of ExecuteDelete can leave the change tracker in a corrupted state where it attempts to
+        // track rows that no longer exist. This can cause an error if you attempt to import a save on first
+        // login, for example.
+        // Since nothing in the tracker is going to be relevant in the face of total deletion, clear the
+        // tracker entirely.
+        this.apiContext.ChangeTracker.Clear();
+
```
SapiensAnatis, line 445: Potentially closes #668 from limited testing, but we'll see

PR #1136, opened 2024-10-30
Faster tests experiment

SapiensAnatis:
Looks like CreateClient runs `Program.cs` every time.

The integration take between 5-7 minutes (ish) now, what does that look like if we re-use the same instance of `HttpClient` across the entire test run?

This will be a gigantic pain in the ass due to DefaultRequestHeaders being shared but could be worth it depending on results


SapiensAnatis:
Saves a couple of minutes...

SapiensAnatis:
Closing in favour of #1170. They aren't mutually exclusive but reusing the client in this way is a gigantic pain in the arse so I'd rather get the savings some other way

PR #1169, opened 2024-11-29
Resolve issues with upgrading dragons/adventurers not being required for endeavors

Alhifar:
Created new MissionCompleteTypes to give a count of dragons/adventurers leveled to a particular level.


DragaliaAPI/DragaliaAPI/Features/Missions/InitialProgress/MissionInitialProgressionService.cs
```diff
@@ -284,6 +292,27 @@ await apiContext
...
+    {
+        if (element != null)
+        {
+            return (
+                await apiContext
+                    .PlayerCharaData.Select(x => new { x.CharaId, x.Level })
+                    .ToListAsync()
+            )
+                .Where(x => MasterAsset.CharaData[x.CharaId].ElementalType == element)
+                .Where(x => x.Level >= level)
```
SapiensAnatis, line 305: Could we bring the level condition into the query to reduce the number of rows we are retrieving?

SapiensAnatis (commented):
This looks pretty good! Thank you for working on it. Just the one minor comment, everything else looks fine to me :)

DragaliaAPI/DragaliaAPI/Features/Missions/InitialProgress/MissionInitialProgressionService.cs
Alhifar: Not super confident with LINQ for SQL in particular, but I think the commit I just added should move the condition into the sql query

SapiensAnatis: Yeah what you've done is correct - anything before the .ToListAsync() will be used to compose the SQL query.

I noticed a few other improvements that could be made. 

- Using `.Count()` instead of `Where.Count()` is more concise, and
- in the no-element case `.ToListAsync().Count` is fetching all the row data when a `.CountAsync()` i.e. SQL `COUNT(*)` will suffice

Just pushed a commit to make those changes rather than do another round of feedback :)

SapiensAnatis (approved):
Thank you!

PR #1170, opened 2024-12-01
Parallelize integration tests

SapiensAnatis:
Use AssemblyFixture to run the integration tests in parallel, rather than in series with a database reset in-between. This allows the tests to complete a fair bit faster.  
Dawnshard is effectively a multi-tenant system, so in theory each test can run in isolation under its own viewer ID without affecting others. The only exceptions are things that are public information, like the news data on the website and time attack rankings.

- Don't set AssertionOptions globally as this appears to trigger exceptions inside FluentAssertions if done from multiple threads
- Various changes to make the tests not interfere with each other, typically by scoping queries and updates in tests to the current viewer ID


SapiensAnatis:
6:16 (random recent test run) to 3:50, pretty good

SapiensAnatis:
There may still be some flaky tests relying on an empty DB and/or mutating global state. I can't vouch I've fixed all of them, we'll just have to fix random failures as they appear.

PR #1180, opened 2024-12-13
Fix off-by-one error when completing Royal Regiment level mission

Alhifar:
Focused on the harder case of overleveling when implementing these, forgot to add a test for hitting the level exactly


SapiensAnatis (approved):
Thanks for looking at this!

PR #1373, opened 2025-08-08
User impersonation page

SapiensAnatis:
Adds an admin-only page that allows access to the existing user impersonation functionality. Previously this had to be accessed via the GraphQL API which was highly inconvenient and unintuitive.

The page is deliberately quite spartan as it is designed only for admins / devs to use.

Access to it and the associated endpoints is driven by the presently unused 'IsAdmin' column in the Players table. Server administrators must manually edit rows in the database to mark them as admins and allow them to access this page.


SapiensAnatis:
No snapshot update: it looks like the sidebar changes are just small enough to fit in screenshot tolerance, and even if I try to update the snapshots, Playwright will do nothing as all the tests pass.

PR #1513, opened 2026-07-04
Route HTTP request logs to test output

SapiensAnatis:
Our integration tests spin up a single HTTP server and send parallel requests to it as different users, for performance reasons. This unfortunately makes it hard to associate logs to any particular test run: parallelism means that if you just capture stdout, you will start to see logs from other tests that ran at the same time.

This is a little experiment to see if we can make a 'smart' logger that uses a HTTP header to redirect output to the appropriate test.


SapiensAnatis:
@claude review these changes

